PR #1, opened 2023-03-04
add postgres via docker, sequelize, migrations for basic user model

timmyichen:
Adds:
* A docker stack to handle dependencies (just postgres for now)
* NPM commands for docker-related things
* Sequelize + DB connection
* Basic `Users` model


.sequelizerc
```diff
@@ -0,0 +1,7 @@
+
+const path = require('path');
+
+module.exports = {
+  'config': path.resolve('server', 'config', 'config.json'),
+  'migrations-path': path.resolve('server', 'migrations')
+};
```
timmyichen, line 7: More docs on [sequelizerc](https://sequelize.org/docs/v6/other-topics/migrations/#the-sequelizerc-file). Normally it defaults to root-level directories, this allows us to specify the directory that contains our config/migrations.

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: "3.9"
```
timmyichen, line 1: We're using docker-compose to run multiple dependencies as docker images locally with a single command. Right now it's a single dependency (postgres) but in the future if we have more, they will be added here.

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
...
+services:
+  pg:
```
timmyichen, line 3: this is the name of the service we're adding

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
...
+    image: postgres:15.2
```
timmyichen, line 4: The image gets pulled from Docker Hub, see [here](https://hub.docker.com/_/postgres)

package.json
```diff
@@ -6,6 +6,8 @@
   "scripts": {
     "dev:server": "cross-env NODE_ENV=development npm run ts:exec server/index.ts",
     "dev": "nodemon",
+    "docker:db": "docker-compose exec pg psql -U postgres -d dev-directory",
```
timmyichen, line 9: This command runs the command `psql -U postgres -d dev-directory` (which just connects to the local db) _on_ the docker container by name, `pg`

server/lib/db.ts
```diff
@@ -1,18 +1,22 @@
...
   constructor(dbName = 'postgres') {
     this.dbName = dbName;
   }
 
-  // Connect to the mongoDB server
+  // Connect to the postgres db
   async connect() {
     // connnect
-    log('Connected to MongoDB! :)');
+    await this.sequelize.authenticate();
```
timmyichen, line 18: This is how we actually connect - [docs](https://sequelize.org/docs/v6/getting-started/#connecting-to-a-database)

server/migrations/20230304021839-create-user.js
```diff
@@ -0,0 +1,33 @@
...
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+    });
+  },
+  async down(queryInterface, Sequelize) {
+    await queryInterface.dropTable('users');
+  }
+};
```
timmyichen, line 33: This is a migration that adds the user table.  See docs on [sequelize migrations](https://sequelize.org/docs/v6/other-topics/migrations/#migration-skeleton).

server/migrations/20230304023314-unique-discorduserid.js
```diff
@@ -0,0 +1,20 @@
...
+  },
+
+  async down (queryInterface, Sequelize) {
+    queryInterface.changeColumn('users', 'discordUserId', {
+      type: Sequelize.STRING,
+      allowNull: false,
+      unique: false,
+    });
+  }
+};
```
timmyichen, line 20: Changes to the schema require multiple migrations (you cannot re-run a previous migration without potentially destroying data).  This change ensures that the `users.discord_user_id` is unique.

server/models/User.model.ts
```diff
@@ -0,0 +1,33 @@
...
+  },
+  discordUserId: {
+    type: DataTypes.STRING,
+    allowNull: false,
+    unique: true,
+  }
+}, {
+  sequelize: Db.sequelize,
+  tableName: 'users',
+});
```
timmyichen, line 31: This is a model definition, allow us to CRUD Users

PR #6, opened 2023-03-09
add auth

timmyichen:
This PR does a bunch of things:
* Adds discord OAuth via jwts (no passport)
* Uses [envalid](https://github.com/af/envalid#readme) to validate environment variables / convert them to the right types
* Uses [express-response-errors](https://www.npmjs.com/package/express-response-errors) to make error-throwing within controllers nicer
* A bunch of test setup stuff - see comments
* Removes the sample router

Look to the comments for more in-depth explanations of things.  Please feel free to ask questions.

You can test this by pulling it down locally and:

Install deps:
```
npm ci
```

Run tests:
```
npm run test:server
```


.env.development
```diff
@@ -1,2 +1,8 @@
 AUTH_SECRET=secret
-REDIRECT_URI=http://localhost:3000/auth-redirect
+DB_PORT=5432
+DB_USER=postgres
+DB_HOST=localhost
+DB_NAME=dev-directory
```
timmyichen, line 5: The DB_STRING environment variable was split up into different pieces to make configuration easier (e.g. for tests, we use all the same values except `DB_NAME=test`.  The `DB_PASSWORD` is the only thing not in here because it's the only thing potentially sensitive.

.env.development
```diff
@@ -1,2 +1,8 @@
...
+DISCORD_GUILD_ID=735923219315425401
+DISCORD_REDIRECT_URI=http://localhost:3000/auth-redirect
+NEXT_PUBLIC_DISCORD_REDIRECT_URI=http://localhost:3000/auth-redirect
```
timmyichen, line 8: Environment variables don't typically get loaded into the client because it's not safe.  We can prefix them with `NEXT_PUBLIC_` to make our client bundler load them in.  See more [here](https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser).

client/lib/oauth.ts
```diff
@@ -0,0 +1,9 @@
+export function getDiscordOauthUrl() {
+  const url = 'https://discord.com/oauth2/authorize';
+  const params = new URLSearchParams();
+  params.append('redirect_uri', process.env.NEXT_PUBLIC_DISCORD_REDIRECT_URI);
+  params.append('response_type', 'code');
+  params.append('scope', 'identify email guilds');
+  params.append('client_id', process.env.NEXT_PUBLIC_DISCORD_CLIENT_ID);
+  return `${url}?${params.toString()}`;
+}
```
timmyichen, line 9: This file builds the discord OAuth URL using the built-in [URLSearchParams API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

package.json
```diff
@@ -39,21 +39,26 @@
     "discord-oauth2": "^2.11.0",
     "dotenv": "^16.0.3",
     "dotenv-flow": "^3.2.0",
+    "envalid": "^7.3.1",
```
timmyichen, line 42: This is used to validate environment variables.  It's nice to use so that if we're missing an environment variable, we'll know immediately (because it will error) instead of hitting a weird error because some value is `undefined`.  More info [here](https://github.com/af/envalid#readme)

package.json
```diff
@@ -39,21 +39,26 @@
...
     "express": "^4.18.2",
     "express-async-router": "^0.1.15",
+    "express-response-errors": "^1.0.5",
```
timmyichen, line 45: [npm link](https://www.npmjs.com/package/express-response-errors)

This lets us do something like the following:
```js
throw new BadRequestError('missing name');
```

instead of this:
```js
return res.status(400).send({ message: 'missing name' })
```

package.json
```diff
@@ -39,21 +39,26 @@
...
     "jsonwebtoken": "^9.0.0",
     "lodash": "^4.17.21",
     "next": "^12.1.2",
+    "next-absolute-url": "^1.2.2",
```
timmyichen, line 49: We are using nextjs, so we should be taking full advantage of its server-side-rendering capabilities. The [`getInitialProps`](https://nextjs.org/docs/api-reference/data-fetching/get-initial-props) of a page is _isomorphic_, which means that it runs on both the client and the server:
* It runs on the server when the page is loaded fresh
* It runs on the client when doing a client-side transition

This is just a [library](https://www.npmjs.com/package/next-absolute-url) to help format our data-fetching calls.

package.json
```diff
@@ -39,21 +39,26 @@
...
     "pg": "^8.9.0",
     "pg-hstore": "^2.3.4",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "sequelize": "^6.29.0",
     "tsconfig-paths": "^4.1.2",
     "typescript": "^4.9.4",
+    "umzug": "^3.2.1",
```
timmyichen, line 57: This is sequelize's tool for running migrations programmatically.  We need to do this in order to run tests on a fresh database (with migrations). [NPM link](https://github.com/sequelize/umzug#umzug)

package.json
```diff
@@ -39,21 +39,26 @@
...
     "universal-cookie": "^4.0.4"
   },
   "devDependencies": {
+    "@ngneat/falso": "^6.4.0",
```
timmyichen, line 61: This is a [library](https://github.com/ngneat/falso#readme) for generating fake data, for tests (and eventually, seeding).

There was more popular one, `faker.js`, but then its owner went rogue and it was a whole story: https://www.revenera.com/blog/software-composition-analysis/the-story-behind-colors-js-and-faker-js/

pages/auth-redirect.tsx
```diff
@@ -0,0 +1,25 @@
...
+      useAuthCode(String(router.query.code));
+    }
+  }, [router.query.code]);
+
+  return (
+    <div><CircularProgress /></div>
+  );
+};
+
+export default AuthRedirect;
```
timmyichen, line 25: This is the page that users are redirected to after authenticating with Discord - Discord gives us a code, which we then pass to the server to fetch their Discord info.  Once successful, it'll redirect back to the home page.

server/api/auth/auth.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (!guilds.find(guild => guild.id === env.DISCORD_GUILD_ID)) {
+    throw new BadRequestError('You must be a member of 100devs discord to join');
+  }
+
+  const user = await upsertUser(discordUser);
+  const token = createAuthToken(user);
+
+  res.cookie(AUTH_COOKIE_NAME, token, { secure: env.isProd });
+  res.sendStatus(200);
+};
```
timmyichen, line 42: In general, the OAuth flow looks like this:
* User authenticates with Discord
* Discord redirects them to one of our pages with a special code
* We use that code to communicate with Discord to get the information we need (the code is short-lived)
* We do whatever with that information
* We sign a jwt token with a secret, set it as a cookie, then respond to the client.

Then every subsequent request from the user will include that cookie.  For information on how it's used, see the `server/middleware/auth.ts` file.

server/api/auth/auth.test.ts
```diff
@@ -0,0 +1,114 @@
...
+});
+
+function mockDiscord(user: Partial<DiscordOauth2.User>, isValid = true ) {
+  return jest.spyOn(authLib, 'getDiscordUserAndGuilds').mockResolvedValueOnce({
+    user: user as DiscordOauth2.User,
+    guilds: [
+      { id: isValid ? env.DISCORD_GUILD_ID : '12345' } as DiscordOauth2.PartialGuild,
+    ],
+  });
+}
```
timmyichen, line 23: We don't actually want to hit discord oauth API with any requests when we're running tests, so this will mock the response.

server/api/users/users.controller.ts
```diff
@@ -0,0 +1,16 @@
...
+import { User } from 'server/models';
+
+type ClientUser = Pick<User, 'id' | 'discord_user_id'>
+
+export const getCurrentUser: RequestHandler<void, ClientUser>  = (req, res) => {
+  // eslint-disable-next-line  @typescript-eslint/no-non-null-assertion
+  const filteredUser = _.pick(req.user!, [
+    'id',
+    'discord_user_id',
+  ]);
```
timmyichen, line 13: We do this filtering so that if we add additional fields to the User model, we aren't including it in the client.  The client in general should know as little as possible in order to prevent unintentional data leakage.

server/lib/auth.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export async function upsertUser(user: DiscordOauth2.User) {
+  const result = await User.upsert({
+    discord_user_id: user.id,
+    email: user.email,
+  });
+
+  return result[0];
+}
```
timmyichen, line 58: In this file, just look at the function names - they're pretty descriptive of what the function does.

server/lib/db.ts
```diff
@@ -1,16 +1,34 @@
...
+    this.dbName = database;
+    this.sequelize = new Sequelize({
+      port: env.DB_PORT,
+      username: env.DB_USER,
+      password: env.DB_PASSWORD,
+      host: env.DB_HOST,
+      database: database,
+      dialect: 'postgres',
+      logging: env.DB_LOGGING,
+    });
```
timmyichen, line 31: Sequelize initialization had to be moved to here since it was dependent on the DB name.

server/middleware/auth.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (validateToken(token)) {
+    const payload = jwt.decode(token) as jwt.JwtPayload;
+    if (payload.user_id) {
+      const user = await User.findByPk(payload.user_id);
+      req.user = user;
+    }
+  }
+
+  next();
+};
```
timmyichen, line 21: This kinda does what you'd expect passport to do - where if the user is logged in, it will check the cookies for a valid token, look up the user from our DB, then attach that user to `req.user`.  No need to figure out serialization/deserialization here!  Though this is definitely the "deserialization" part (the serialization happens when we sign the token when they login)

server/middleware/auth.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const requireUser: RequestHandler = async (req, res, next) => {
+  if (!req.user) {
+    throw new UnauthorizedError();
+  }
+
+  next();
+};
```
timmyichen, line 29: You'd use this middleware on routes that you want to be only accessible by logged-in users.

server/models/index.ts
```diff
@@ -0,0 +1,5 @@
+import User from './User.model';
+
+export {
+  User,
+};
```
timmyichen, line 5: A single entry point for all models.  All models should be exported from here.  This way, instead of doing something like the following:

```js
import ModelA from 'server/models/ModelA.model.ts';
import ModelB from 'server/models/ModelB.model.ts';
import ModelC from 'server/models/ModelC.model.ts';
```

It would just be:

```js
import {
  ModelA,
  ModelB,
  ModelC,
} from 'server/models';
```

server/server.ts
```diff
@@ -23,12 +29,17 @@ class Server {
     this.app.use(express.json());
     this.app.use(express.urlencoded({ extended: true }));
     this.app.use(cookieParser());
+    this.app.use(attachUser);
```
timmyichen, line 32: We want to make sure this happens _after_ we parse cookies, but _before_ we handle any of the API routes, since those might be dependent on `req.user`.

server/test/server.ts
```diff
@@ -1,32 +1,109 @@
...
     });
   }
+
+  login(user: User) {
+    this.loggedInUser = user;
+  }
+
+  logout() {
+    this.loggedInUser = null;
+  }
```
timmyichen, line 96: All of this is so that we can "pretend" to log in as users; otherwise any authed flow would be difficult to test since we're not actually hitting the discord API.

server/test/server.ts
```diff
@@ -1,32 +1,109 @@
...
+
+  async destroy() {
+    await this.revertMigrations();
+    await this.db.sequelize.close();
+    if (this.server) {
+      return new Promise<void>((resolve) => {
+        this.server.close(() => resolve());
+      });
+    }
+  }
```
timmyichen, line 106: When we take down the test server, we also want to revert migrations, which will delete all the data from the test DB.  This makes it ready to be used for the next test without any unintentional side effects of leftover data.

server/test/server.ts
```diff
@@ -1,32 +1,109 @@
...
+
+  async runMigrations() {
+    await this.umzug.up();
+  }
+
+  async revertMigrations() {
+    await this.umzug.down({ to: 0 as const });
+  }
```
timmyichen, line 75: The test database must be created; it is not created as a part of the docker container setup, so we do it here in the tests.  Migrations are also required for us to get the tables we're expecting, so we need some methods to stand run migrations and undo migrations programmatically.

server/test/server.ts
```diff
@@ -1,32 +1,109 @@
...
+        resolve: ({ name, path, context }) => {
+          // eslint-disable-next-line @typescript-eslint/no-var-requires
+          const migration = require(path);
+          return {
+            name,
+            up: async () => migration.up(context, Sequelize),
+            down: async () => migration.down(context, Sequelize),
+          };
+        },
+      },
```
timmyichen, line 40: this is annoying to do but necessary - see the link in the comment.

wo1vin (approved):
I'm unfamiliar with a lot of the packages, tech, and code referenced, but I am getting to understand how it all works together based on your comments.

jmoldyvan:
nice tim!

.env.development
MatthewBozin: This is a great change, would you recommend we use this convention in Together / Binary Upload Zoom as well?

code2244 (approved):
This is great, all working for me!

MatthewBozin (approved):
Looks good! Really appreciate the explanatory comments!

.env.development
devonzara: Bit late, as I see it was merged... but, since the trailing path on the `REDIRECT_URI`s is fixed and won't change per installation, could we simplify the env by replacing them with a single, more generic `BASE_URL`?
```js
`${process.env.BASE_URL}/auth-redirect`
```

Client-side we can create a simple function to build it using `window.locaion`. eg.
```js
const baseUrl = () => {
    const { protocol, hostname, port } = window.location;
    return `${protocol}//${hostname}${port && `:${port}`}`;
};
```

Also, perhaps a more conventional endpoint URI instead of `/auth-redirect`, eg.
```js
`${baseUrl()}/auth/callback`

// or auth/[provider]/callback
`${baseUrl()}/auth/discord/callback`
```

timmyichen: Good points! I would agree if this weren't also a config that we have to set on the the Discord developer dashboard: 
image

Given that, I think it's generally easier to keep as one big string rather than a computed value to keep it in sync with the discord dashboard

devonzara: Ahh, that's fair. 👍🏼 I suppose I was viewing it the other way around... as if the code would be the authoritative source of truth and you'd change Discord to match.

Issue #11, opened 2023-03-10
Create simple page for a profile


timmyichen:
Project: #19 
Depends on: #9 

**Description**
There are no designs yet for this, so just do whatever's simplest and looks just okay.  When a user navigates to `/profile/:id`, they should be shown that user's profile.  Use the newly created route in #9 to fetch the data.  No need to display the user's internal `id`.  Use the user's `discord_name` as the main heading.

The user's profile should be server-side-rendered, so there should be no loading state.

You'll probably have to add a new server-side router to handle page routes; unless there's a better nextjs-way of doing this when the page name is also a parameter.

**Acceptance Criteria**
* The page exists and is server-side rendered
* The page displays all of the user's information returned from the GET endpoint

devonzara:
Initial groundwork is done... blocked on #8.

PR #44, opened 2023-03-16
feat: add initial profile pages

devonzara:
Closes:
 - #11 

# Description

Creates a simple profile page that displays the user's Discord name, bio, and links to their social media accounts.

## Testing

Tests needed

## Type of change

## Screenshots

image

# Checklist:

### DIFF
--- a/pages/profile/[id].tsx
+++ b/pages/profile/[id].tsx
@@ -16,10 +16,6 @@ type Props = {
 };
 
 const ProfilePage: NextPage<Props> = ({ user }: Props) => {
-  if (!user) {
-    return <div>User not found</div>;
-  }
-
   function possessiveForm(name: string): string {
     return name.endsWith('s') ? `${name}'` : `${name}'s`;
   }
@@ -112,7 +108,7 @@ export const getServerSideProps: GetServerSideProps<Props> = async ({ req, query
 
     return { props: { user } };
   } catch (error) {
-    return { props: { notFound: true } };
+    return { notFound: true };
   }
 };
 
--- a/pages/profile/[id].tsx
+++ b/pages/profile/[id].tsx
@@ -98,7 +98,7 @@ const ProfilePage: NextPage<Props> = ({ user, statusCode }: Props) => {
   );
 };
 
-ProfilePage.getInitialProps = async ({ req, res, query }: NextPageContext): Promise<Props> => {
+ProfilePage.getInitialProps = async ({ req, res, query }: NextPageContext) => {
   const props: Props = {
     user: null,
     statusCode: 200,
--- a/pages/profile/[id].tsx
+++ b/pages/profile/[id].tsx
@@ -3,20 +3,25 @@ import HomeRoundedIcon from '@mui/icons-material/HomeRounded';
 import LinkedInIcon from '@mui/icons-material/LinkedIn';
 import TwitterIcon from '@mui/icons-material/Twitter';
 import { Container, Box, ListItemIcon, ListItemButton, List, Typography } from '@mui/material';
-import axios, { RawAxiosRequestConfig } from 'axios';
-import { GetServerSideProps, NextPage } from 'next';
-import absoluteUrl from 'next-absolute-url';
+import { NextPage, NextPageContext } from 'next';
+import ErrorPage from 'next/error';
 import React from 'react';
 
+import createAxiosInstance from 'client/lib/axios';
 import { UserProfile } from 'server/types/User';
 
-type Props = {
+interface Props {
   user?: UserProfile;
-};
+  notFound?: boolean;
+}
 
-const ProfilePage: NextPage<Props> = ({ user }: Props) => {
+const ProfilePage: NextPage<Props> = ({ user, notFound = null }: Props) => {
   function possessiveForm(name: string): string {
     return name.endsWith('s') ? `${name}'` : `${name}'s`;
+  }
+
+  if (notFound === true) {
+    return <ErrorPage statusCode={404}/>;
   }
 
   return (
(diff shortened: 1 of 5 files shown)


pages/profile/[id].tsx
```diff
@@ -0,0 +1,114 @@
...
+              </ListItemIcon>
+            </ListItemButton>
+          )}
+        </List>
+      </Container>
+    </>
+  );
+};
+
+export const getServerSideProps: GetServerSideProps<Props> = async ({ req, query }) => {
```
Comment (timmyichen) [inline pages/profile/[id].tsx:95]: can we make this use getInitialProps like we do in the [index page](https://github.com/timmyichen/dev-directory/blob/main/pages/index.tsx#L44)?  for consistency, and this way we save a round trip request to the server

pages/profile/[id].tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  try {
+    const config: RawAxiosRequestConfig = req ? {
+      withCredentials: true,
+      headers: {
+        cookie: req.headers.cookie,
+      },
+    } : {};
+
+    const { data: user } = await axios.get<UserProfile>(`${origin}/api/users/${query.id}`, config);
```
Comment (timmyichen) [inline pages/profile/[id].tsx:106]: how do you feel about doing a minor refactor here - pulling this out into a client lib file for making requests - so that every API request doesn't need to re-define the axios config? :) totally optional, i get if you'd rather it be a separate task (that either you or anyone else can take on)

timmyichen (commented):
looks great! haven't tested it locally yet, but small comment about nextjs bits. thanks for doing this!

pages/profile/[id].tsx
Comment (devonzara) [reply pages/profile/[id].tsx:106]: I actually made a note about that and was going to create an issue or do it myself later in a separate PR, but yea... I can throw it in. 👍🏼

pages/profile/[id].tsx
Comment (devonzara) [reply pages/profile/[id].tsx:95]: Even though it's not recommended? I was thinking we'd change Index to use `getServerSideProps` instead. 🤔

Comment (timmyichen) [reply pages/profile/[id].tsx:95]: yeah, in the [3/10 meeting](https://github.com/timmyichen/dev-directory/discussions/20) we decided to go with a REST API (which implicitly means we're using `getInitialProps`. `getServerSideProps` always runs on the server; next just makes a secret endpoint to fetch that data.  the flow would look like this:

getInitialProps:
* server side render: server -> makes request to endpoint -> renders -> sends to client
* client side render: client -> makes request to endpoint -> gets response

getServerSideProps:
* server side render: server -> makes request to endpoint -> renders -> sends to client
* client side render: client -> makes request to server -> server makes request to endpoint -> server receives data -> server passes data back to client

so we'd be making an extra fetch on the user side

the way that `getServerSideProps` is recommended is to do all the _direct_ data fetching from that function, rather than to call our own server's endpoint to get the data

[here](https://github.com/vercel/next.js/discussions/10930)'s a relevant thread if you want to give it a read!

pages/profile/[id].tsx
```diff
@@ -92,21 +97,14 @@ const ProfilePage: NextPage<Props> = ({ user }: Props) => {
...
-
-    const { data: user } = await axios.get<UserProfile>(`${origin}/api/users/${query.id}`, config);
+    const axios = createAxiosInstance(req);
+    const { data: user } = await axios.get<UserProfile>(`/api/users/${query.id}`);
 
-    return { props: { user } };
+    return { user };
   } catch (error) {
+    res.statusCode = error.response.status;
     return { notFound: true };
```
Comment (timmyichen) [inline pages/profile/[id].tsx:108]: should we instead return a null user if not found?  that way we can basically use the truthiness of `user` as whether the user was found or not, and we wouldn't need the optional chaining e.g. `user?.website`

timmyichen (commented):
one small comment, otherwise LGTM!

pages/profile/[id].tsx
Comment (devonzara) [reply pages/profile/[id].tsx:108]: Honestly, the optional chaining should never happen because it'll show the 404 template instead.

I just put the optional chaining there while debugging/manual testing and left them because they don't hurt anything. They can be removed if we'd prefer that.

Comment (timmyichen) [reply pages/profile/[id].tsx:108]: yeah, i guess what i'm saying is that if we use `user` as the proxy for `notFound`, we can do something like this:

```js
if (!user) {
  return <NotFound />
}

// user guaranteed to be truthy here
return (
  <div>{user.username}</div>
)
```

Comment (devonzara) [reply pages/profile/[id].tsx:108]: Ahh, yea. We can do that. I just kept it the same as what `getServerSideProps` expected.

Comment (devonzara) [reply pages/profile/[id].tsx:108]: Actually, thoughts on passing the actual `statusCode`? Since errors other than just 404 may occur.

```js
if (statusCode !== 200) {
  return <ErrorPage statusCode={statusCode}/>;
}
```

We can still pass a null `user`...
```js
getInitialProps... {
  const props: Props = {
    user: null,
    statusCode: 200,
  };

  try {} catch() {}

  return props;
}
```

or default it to `null` in the params...
```js
const ProfilePage: NextPage<Props> = ({ user = null, statusCode }: Props) => {
...
```

Also, did you still want the optional chaining removed?

pages/profile/[id].tsx
```diff
@@ -0,0 +1,118 @@
...
+              </ListItemIcon>
+            </ListItemButton>
+          )}
+        </List>
+      </Container>
+    </>
+  );
+};
+
+ProfilePage.getInitialProps = async ({ req, res, query }: NextPageContext): Promise<Props> => {
```
Comment (timmyichen) [inline pages/profile/[id].tsx:101]: does the return type need to be explicitly defined here?

timmyichen (approved):
Thanks for making the changes! Just one small thing but non-blocking

pages/profile/[id].tsx
Comment (devonzara) [reply pages/profile/[id].tsx:101]: Need to be, no. Should it be? Depends what team you're on. Some say always, some say never, and others say it depends... are you branching? 🤔

Comment (devonzara) [reply pages/profile/[id].tsx:101]: Fair. The jury is still out for me. I'm leaning towards always as I feel like being explicit is a nice reminder, especially when refactoring. Anyway, I can remove it.

Caleb-Cohen:
Hello! How should we be testing this?

When running this PR in local I am not able to re-create exactly what @devonzara has completed.

Screenshot 2023-03-22 at 10 43 14 AM

devonzara:
@Caleb-Cohen If you can view the profile page with your name, it's working. There isn't a form to populate the other profile fields so you'd have to add them to the database manually via CLI or a GUI like TablePlus or DBeaver.

Caleb-Cohen (approved):

DevinCLane:
Not sure I follow how to test this locally, running the PR and then logging in gives me 

"Unhandled Runtime Error

AxiosError: Request failed with status code 500"


Screenshot 2023-03-22 at 11 53 41 AM

devonzara:
@DevinCLane let me know when you're available and we can jump in a call together.

DevinCLane:
thanks for fixing that! @devonzara

PR #48, opened 2023-03-18
#38 create test utilities

### LINKED ISSUE #38: Create test utilities
**Related Issues**
Depends on: #9

**Description**
Most of our tests will depend on users being created.  Creating a full-fledged users with all the fields is cumbersome, so we can make it a bit easier by adding some test utilities for creating a random user.

To this end, we'll create a new file `server/test/utils.ts`, which exports a new function `createUser` which creates a user with randomly generated fields in all fields.  The arguments should just be a subset of User fields (excluding ID, but this doesn't have to be strict since it's just used for testing), where any fields specified are used to create the user, and any unspecified fields are randomly generated using [falso](https://github.com/ngneat/falso).

For example, the following:

```
createUser({ discord_name: "poop#1234", bio: "i like poop" })
```

Should create a user that has their `discord_name` and `bio` to be the above values, and random values for everything else.  Bonus points if there's some low percentage (like 20%) that a non-required field is blank.

**Acceptance Criteria**
* The utility function exists and is used in tests instead of creating users manually

MainlyColors:
Closes #38 

# Description

- Add `server/test/utils.ts`
  - purpose to create a `createUser` function that uses falso.js to mockup fake users and creates them in the DB
  - 20% random chance for any field besides `id` to be an empty string
  - function also accepts input for any of the fields besides `id` to be manually input, they won't be wiped out by the randomEmptyString function
 
- Replaced `User.create()` with new `createUser()` function in `server\api\users\users.test.ts` & `server\api\auth\auth.test.ts`

*no new packages*
## Testing

Describe what a reviewer should do to test your code, including instructions for manually testing as well as relevant automated test suites to run.

- `npm run test:server`

*No manual testing required*

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Screenshots

passing tests
image

# Checklist:


devonzara (changes requested):

server/api/auth/auth.test.ts
```diff
@@ -80,11 +81,7 @@ describe('auth router', () => {
       const oldEmail = randEmail();
       const newEmail = randEmail();
 
-      await User.create({
-        discord_user_id: id,
-        discord_name: name,
-        email: oldEmail,
-      });
+      await createUser();
```
timmyichen, line 84: aha! so - even though tests pass - i don't think it's necessarily testing the same thing.  we need to create the email using the replaced code's deets, otherwise the first assertion (on `userWithOldEmail` will always be `null` because the user is never created, which also makes the second assertion kinda pointless since it's guaranteed by the mocked discord value

server/test/utils.ts
```diff
@@ -0,0 +1,94 @@
+import { rand, randEmail, randNumber, randUserName, randQuote, randBrand } from '@ngneat/falso';
+
+import { User } from 'server/models';
+
+// user interface
```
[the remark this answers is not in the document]
timmyichen: in this case i think they can/should be optional, since it will fall back to random values if not set.  e.g. "i want a user with this specific email but i dont care about the other values"

server/test/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  website?: string;
+}
+
+// https://stackoverflow.com/questions/39494689/is-it-possible-to-restrict-number-to-a-certain-range/39495173#39495173
+// Tim's fault for telling me this is possible
+type Enumerate<N extends number, Acc extends number[] = []> = Acc['length'] extends N
+  ? Acc[number]
+  : Enumerate<N, [...Acc, Acc['length']]>;
+
+type IntRange<F extends number, T extends number> = Exclude<Enumerate<T>, Enumerate<F>>;
```
timmyichen, line 23: wheee 🙃 i think these are probably best in a separate types file, since it's pretty general-purpose and may be used for other things

timmyichen: maybe `server/types/utils.ts`?

server/test/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+
+function randWebsite() {
+  const scheme = rand(['https', 'http']);
+  const wwwPart = rand(['www.', '']);
+  const regex = new RegExp('[\\W]', 'g');
+  const host = randBrand().replace(regex, '');
+  const tld = rand(['com', 'gg', 'dev', 'org', 'net', 'me', 'io', 'us', 'info']);
+
+  return `${scheme}://${wwwPart}${host}.${tld}`;
+}
```
timmyichen, line 60: we can use falso for this: https://ngneat.github.io/falso/docs/internet#randurl

server/test/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+    randNumber({
+      min: 0,
+      max: 9999,
+    })
+  ).padStart(4, '0')}`;
+}
+
+function randUserNameHandle() {
+  return `@${randUserName({})}`;
+}
```
timmyichen, line 46: i think we actually _don't_ want the `@` included; it's not validated but we shouldn't be accepting special characters in the username anyway.  Basically they should be giving a handle like `timmyichen` and then we can render it as `@timmyichen` - not on the user to supply it

server/test/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+
+// example
+
+function randDiscordUserName() {
+  return `${randUserName({})}#${String(
+    randNumber({
+      min: 0,
+      max: 9999,
+    })
+  ).padStart(4, '0')}`;
```
timmyichen, line 41: i think this part would actually be preferred to be its own function, like `getRandomDiscriminator`

server/test/utils.ts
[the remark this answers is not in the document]
timmyichen: but in any case you can use the `Partial` type for this: https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype

server/test/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  twitter_username,
+  linkedin_url,
+  github_username,
+  website,
+}: UserObject = {}) {
+  // build object
+  // check which properties are includes
+  // 20% chance at empty values
+  const userObject = {
+    email: email ?? randomEmptyChance(20, randEmail()),
```
timmyichen, line 77: certain fields (like email, discord_user_id) should always have a value and should thus never be falsy e.g. an empty string - so these should probably not use the `randomEmptyChance` function

timmyichen (commented):
Great start! I have a few comments, plus one general comment:

Think of functions as actions - you're _doing_ something when you're calling a function, and thus generally they should be named as actions - with verbs.  So instead of `linkedInURL`, you might call it `getLinkedInUrl` instead (and might even want to accept the username as an argument as this will likely be useful on the client as well).

server/test/utils.ts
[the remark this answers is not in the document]
devonzara: > in this case i think they can/should be optional, since it will fall back to random values if not set. e.g. "i want a user with this specific email but i dont care about the other values"

Ahh, you're right... I missed that it was being used for incoming params and confused it with `userObject`, which isn't explicitly given a type. 👍🏼

server/test/utils.ts
MainlyColors: I'm cool with randurl and didn't see it before but the way it's implemented doesn't give a chance for 'www' to show up.

Will that create a blind spot for testing logic and UI?

from https://github.com/ngneat/falso/blob/main/packages/falso/src/lib/url.ts
```ts
 return `${fake(['http', 'https'])}://${randWord()}.${randDomainSuffix()}`;
```

timmyichen: nope! `www` is not required for URLs anyway

timmyichen (approved):
minor comment about using falso for the URL still stands, but not blocking on it - otherwise LGTM!

server/test/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  : Enumerate<N, [...Acc, Acc['length']]>;
+
+type IntRange<F extends number, T extends number> = Exclude<Enumerate<T>, Enumerate<F>>;
+
+// random
+// probability should only be 1 - 100
+function randomEmptyChance(probability: IntRange<0, 100>, preferredResult: string) {
+  // out of a 100
+  const randomNum = Math.random() * 100;
+  return randomNum < probability ? '' : preferredResult;
```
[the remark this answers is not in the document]
Caleb-Cohen: What is overriding defaults?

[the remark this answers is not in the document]
devonzara: @Caleb-Cohen Overriding defaults means changing default behavior.

For instance, the `createUser` function accepts an object with each `User` field as a possible key. If a key doesn't exist in the object (or if no object is provided), it will generate a random values with the possibility of some fields being empty. That random value or empty would be the default behavior... passing in a specific value would be an override.

However, because `randomEmptyChance` isn't exported, a user can't pass in an override and maintain the possibility of it being empty. Probably not something you'd need often, but being able to replicate that empty chance from outside the utility functions would be useful. eg.
```js
// Extra long discriminator (unrealisitc example)
const username = `${randUser({ length: 10 })}#randNumber({ length: 6 });

createUser({
    // Retains the possibility of being empty
    discord_name: randomEmptyChance(20, username),
});
```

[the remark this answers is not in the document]
MainlyColors: This function is overriding defaults by giving a low chance (20% ) to return an empty string otherwise it will return the passed-in value, for all cases this value is another random function provided by falso for each related property. 

This function is being used on the userObject that is built by `createUser`

https://github.com/timmyichen/dev-directory/blob/d45bf4c12d28ce68af59adbab0ad0c3d152572df/server/test/utils.ts#L52-L61


An example of this with `twitter_username` 
(only the single property is shown for example purposes, but normally all the properties would be there) 

```ts
twitter_username: twitter_username ?? randomEmptyChance(20, randUserName())

// createUser({twitter_username: 'example'})
// output userObject = {twitter_username: 'example'}

// createUser()
// output userObject = {twitter_username: 'Joyce_Azulay85'}

// createUser()
// output if randomEmptyChance triggers userObject = {twitter_username: ''}
```

devonzara (approved):

PR #49, opened 2023-03-18
chore: rearrange .env files

devonzara:
Closes N/A

# Description

We were operating under the understanding that the `.env.<environment>` files set the defaults for each environment and the `.env` file would override. Based on [dotenv-flow's documentation](https://github.com/kerimdzhanov/dotenv-flow#files-under-version-control), that is the opposite of how it works.

This change introduces a `.env.defaults` to initialize default values, users are still instructed to create a `.env` for configuring secrets, and the environment-specific `.env.<environment>` files will have the highest priority and overwrite all other files.

## Testing

Ensure that the project starts and the tests still pass.

## Type of change

## Screenshots

N/A

# Checklist:


timmyichen (approved):
Thanks for making this change!

.env.sample
```diff
@@ -1,4 +1,6 @@
-DISCORD_CLIENT_ID=1082239786162606080
-NEXT_PUBLIC_DISCORD_CLIENT_ID=1082239786162606080
-DISCORD_CLIENT_SECRET=123
-DB_PASSWORD=password
+# Secrets
+AUTH_SECRET=secret
+
+DISCORD_CLIENT_SECRET=
```
MainlyColors, line 4: Should the DISCORD_CLIENT_SECRET be filled out here since it was filled out before? AUTH_SECRET is still filled out

devonzara: It's going to fail the same being empty as it does as `123` and I believe leaving it blank is more indicative that it needs to be filled in for certain features to work.

`AUTH_SECRET` is used when generating session tokens. It could probably be left empty too as I believe `envalid` would convert it to an empty string. However, it's mainly only important on prod so setting a default for development isn't a concern.

devonzara: Updated all default secrets to be empty, populated the sample env with common development values.

MainlyColors: That makes sense now, I didn't know what a client secret was before. thought it worked the same way AUTH_SECRET did.

MainlyColors (approved):
LGTM

Documenting Thoughts:
Thought we needed `docker-compose.test.yml` file but looks like when no `-f` option is given to `docker-compose` then docker will look for the default `docker-compose.yml` file and we were already creating a new test DB with the `createDb()` method in `server/test/server.ts` `TestServer` class so there is no reason to have separate `docker-compose.test.yml` file as it wasn't being used

timmyichen:
> Documenting Thoughts: Thought we needed `docker-compose.test.yml` file but looks like when no `-f` option is given to `docker-compose` then docker will look for the default `docker-compose.yml` file and we were already creating a new test DB with the `createDb()` method in `server/test/server.ts` `TestServer` class so there is no reason to have separate `docker-compose.test.yml` file as it wasn't being used

ah it's being used in the github action [here](https://github.com/timmyichen/dev-directory/blob/main/.github/workflows/lint.yml#L43)

I think @MatthewBozin might remember better why exactly it was needed

Caleb-Cohen (approved):
LGTM, Devon never does wrong.

Issue #10, opened 2023-03-10
Add API endpoint for updating a user's profile


timmyichen:
Project: #19 
Depends on: #8 

**Description**
This issue adds an API endpoint for updating a user's profile.

Add an API endpoint:
* `PATCH /api/users/:id`

This endpoint should allow the updating of data for the user's model, limited to the following fields:
* `bio`
* social links added in #8 

The endpoint should throw a ForbiddenError if it's called on any user except for the currently-authed user (i.e. people can only edit their own profiles)

Partial updates should be allowed; i.e. not every field above is required when calling this endpoint.

User-submitted data should be sanitized to not include markup (e.g. can't set your `bio` to be `<script>alert('hi')</script>`)

**Acceptance Criteria**
* API endpoint exists
* New acceptance tests exist and pass (covering edge cases)
* Existing tests pass

PR #50, opened 2023-03-19
#10 api endpoint update user profile

MatthewBozin:
Closes #issue-number-here #10

# Description

- Added PATCH endpoint to user api that modifies user data for specific fields.
- Added middleware that throws error if user tries to modify data that is not theirs.
- Added middleware that removes markup from fields in req.body. This uses the isomorphic-dompurify npm library.
- Added acceptance tests for the new endpoint and middleware.

## Testing

Run `npm run test:server`

## Type of change

# Checklist:


server/middleware/users.ts
```diff
@@ -0,0 +1,21 @@
+import { ForbiddenError } from 'express-response-errors';
+import DOMpurify from 'isomorphic-dompurify';
+
+import { AnyRequestHandler } from 'server/types/express';
+
+export const requireSameId: AnyRequestHandler = async (req, res, next) => {
+  if (req.user.dataValues.id.toString() !== req.params.id) {
+    throw new ForbiddenError('User can only make changes to their own profile');
+  }
```
timmyichen, line 9: just curious, if you had done `req.user.id.toString()` would that have worked as well?

server/api/users/users.controller.ts
```diff
@@ -18,15 +18,23 @@ export const getCurrentUser: RequestHandler<void, ClientUser>  = (req, res) => {
...
+  | 'github_username'
+  | 'website'>
+
+interface UpdatableFields {
+  bio?: string;
+  twitter_username?: string;
+  linkedin_url?: string;
+  github_username?: string;
+  webiste?: string;
+}
```
timmyichen, line 35: for this you can use a combination of built-in typescript `Partial` and `Pick`

also, typo for website :)

server/api/users/users.test.ts
```diff
@@ -72,4 +72,96 @@ describe('auth router', () => {
...
+
+      const fieldsToUpdate = {
+        bio: 'I was updated.',
+        twitter_username: 'Updated@updated',
+        linkedin_url: 'updated',
+        github_username: 'gotUpdated',
+        website: 'https://thiswasupdated.com/',
+      };
+
+      const res = await server.exec.patch(`/api/users/${server.loggedInUser.dataValues.id}`).send(fieldsToUpdate);
```
timmyichen, line 109: instead of `server.loggedInUser.dataValues.id` i think you can just do `user.id` here

server/api/users/users.controller.ts
```diff
@@ -46,3 +54,34 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+    where: {
+      id: req.params.id,
+    },
+  });
+
+  if (updated[0] === 0) {
+    throw new NotFoundError('User not found');
+  }
+
+  res.json(`User of id ${req.params.id} updated.`);
```
timmyichen, line 86: returning a string that basically amounts to "this action was successful" can be replaced by just sending a 200 response with no message (the client can deal with the message that gets surfaced in the UI).  however, we might have more use here to return the entire updated user object itself, so that the client can more easily update what it has stored.

server/api/users/users.test.ts
```diff
@@ -72,4 +72,96 @@ describe('auth router', () => {
...
   });
+
+  describe('PATCH /:id', () => {
+    it('returns 403 if profile id is not equal to logged in user id', async () => {
+      const res = await server.exec.patch('/api/users/54321');
+      expect(res.status).toBe(403);
+    });
+
+    it('returns 404 if specified user doesn\'t exist', async () => {
+      const res = await server.exec.patch('/api/users/1');
```
timmyichen, line 83: wouldn't an id of `1` possibly attempt this action on a user that exists, since user IDs are autoincrementing starting from 1? 😅

server/api/users/users.test.ts
```diff
@@ -72,4 +72,96 @@ describe('auth router', () => {
...
+        id: user.id,
+        discord_user_id: user.discord_user_id,
+        discord_name: user.discord_name,
+        bio: '',
+        twitter_username: 'Updated@updated',
+        linkedin_url: 'updated',
+        github_username: 'gotUpdated',
+        website: 'https://thiswasupdated.com/',
+      });
+    });
```
timmyichen, line 165: i think we can actually do away with this test and instead add tests to the middleware functions!  see next comment

server/middleware/users.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export const removeMarkup: AnyRequestHandler = async (req, res, next) => {
+  const fields = Object.keys(req.body);
+  fields.forEach((field) => {
+    req.body[field] = DOMpurify.sanitize(req.body[field]);
+  });
+
+  next();
+};
```
timmyichen, line 21: we can test middleware too! no server/`TestServer` instance needed, you can just test it as if it were any old function - just pass a (req with a body, empty object res, and an unimplemented [jest mock function](https://jestjs.io/docs/jest-object#jestfnimplementation) for next), call the middleware, and you can verify that `req.body` should've been mutated to account for it + the next function was called

timmyichen (commented):
looking great! left a few comments for optimizations but i think this is a great start

MatthewBozin:
@timmyichen just saw these, email notifs were going to spam D: Will get to these tomorrow!

server/api/users/users.controller.ts
```diff
@@ -46,3 +54,34 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+
+  const toUpdate: UpdatableFields = {};
+
+  acceptableFields.forEach((field) => {
+    if (field in req.body) {
+      toUpdate[field as keyof UpdatableFields] = req.body[field as keyof UpdatableFields];
+    }
+  });
+
+  const updated = await User.update(toUpdate, {
```
devonzara, line 76: How will this handle an empty object? Should a query be executed at all if no fields have changed?

server/middleware/users.test.ts
```diff
@@ -0,0 +1,21 @@
+import { Request, Response } from 'express';
+import { NextFunction } from 'express-async-router';
+
+import { removeMarkup } from './users';
+
+
+describe('removeMarkup middleware', () => {
+  it('should remove markup from request body fields', () => {
```
devonzara, line 8: Should we actually be removing the markup or converting it to HTML entities instead?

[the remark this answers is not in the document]
timmyichen: what do you mean by converting to HTML entities?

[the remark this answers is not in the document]
devonzara: I'm just thinking of things like Twitter bios... people often put HTML tags in them as a form of expression. Should be fine to leave them as long as the chevrons are converted to `&lt;` and `&gt;`, right?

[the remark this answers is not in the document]
timmyichen: ah yeah good callout.  is a find-and-replace-all of `<` and `>` with lt/gt respectively enough to prevent xss? heh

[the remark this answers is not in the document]
devonzara: Not sure, tbh... I'd probably also escape `&`, `'`, and `"` though a library specific to this would probably be the safer option.

[he](https://www.npmjs.com/package/he) seems to be extremely popular (15m downloads weekly), but hasn't had any changes since 2018.
[sanitize-html](https://www.npmjs.com/package/sanitize-html) seems to have more active development, but only 1.3m weekly downloads.

server/api/users/users.controller.ts
MatthewBozin: That's a great question! A query probably shouldn't be executed. I've added a check and test for this!

server/middleware/users.test.ts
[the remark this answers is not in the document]
timmyichen: @MatthewBozin thoughts on using an alternative like this? also let us know if you're super busy with final interview stuff, we can get someone to take this over the finish line

server/api/users/users.controller.ts
```diff
@@ -49,6 +49,10 @@ export const getUserById: RequestHandler<{ id: string }, UserProfile> = async (r
 
 export const updateUserById: RequestHandler<{ id: string }, string, UpdatableFields> = async (req, res) => {
 
+  if (Object.keys(req.body).length === 0) {
+    throw new BadRequestError('Request body should not be an empty object.');
+  }
```
timmyichen, line 54: shouldn't we instead be checking for `toUpdate`? since they could have a req.body.asdf which wouldn't get picked by the update object, but wouldn't be caught here and would result in a no-op update later

server/api/users/users.controller.ts
```diff
@@ -38,3 +46,38 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+    'website',
+  ];
+
+  const toUpdate: UpdatableFields = {};
+
+  acceptableFields.forEach((field) => {
+    if (field in req.body) {
+      toUpdate[field as keyof UpdatableFields] = req.body[field as keyof UpdatableFields];
+    }
+  });
```
timmyichen, line 70: you can use [lodash pick](https://lodash.com/docs/4.17.15#pick) here

server/api/users/users.controller.ts
```diff
@@ -18,7 +18,15 @@ export const getCurrentUser: RequestHandler<void, ClientUser>  = (req, res) => {
...
 };
 
-export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req, res) => {
+interface UpdatableFields {
+  bio?: string;
+  twitter_username?: string;
+  linkedin_url?: string;
+  github_username?: string;
+  website?: string;
+}
```
timmyichen, line 27: nit: let's colocate this with the controller method iself; i.e. put above `updateUserById`

timmyichen (commented):
some minor comments, but otherwise it's looking good! main thing is just that thread i pinged you on with devon

server/api/users/users.test.ts
```diff
@@ -57,4 +60,88 @@ describe('auth router', () => {
...
   });
+
+  describe('PATCH /:id', () => {
+    it('returns 403 if profile id is not equal to logged in user id', async () => {
+      const res = await server.exec.patch('/api/users/54321');
+      expect(res.status).toBe(403);
+    });
+
+    it('returns 400 if request body is empty object', async () => {
+      const user = await User.create({
```
MainlyColors, line 71: We recently merged #48 that includes a `createUser()` method for randomly generating a user that can be used here instead of manually typing it out.

You can see an example of how the method is used for the `'GET /:id'` test further up

server/middleware/users.test.ts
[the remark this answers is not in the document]
MatthewBozin: It looks like sanitize-html's latest build is failing, so he is looking like a better option, plus it seems to have a better set of methods for what we want. I'll look into bringing it in and creating some tests for it

server/api/users/users.test.ts
MatthewBozin: That's great, thanks for the heads up! I've gone ahead and replaced the user creation code with it.

server/api/users/users.controller.ts
```diff
@@ -38,3 +38,43 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+    'linkedin_url',
+    'github_username',
+    'website',
+  ];
+
+  for (const field of fieldsToUpdate) {
+    if (field in req.body) {
+      break;
+    }
+    throw new BadRequestError('Request body should contain at least one updatable field.');
```
devonzara, line 64: I believe this can be simplified using a conditional and `Object.keys(toUpdate).length` or even `_.size(toUpdate)` after the `pick`.

server/api/users/users.test.ts
```diff
@@ -36,4 +37,65 @@ describe('auth router', () => {
...
+      const res = await server.exec.patch('/api/users/54321');
+      expect(res.status).toBe(403);
+    });
+
+    it('returns 400 if request body does not contain any updatable fields', async () => {
+      const user = await createUser();
+
+      server.login(user);
+
+      const res = await server.exec.patch(`/api/users/${user.id}`).send({ asdf: 'asdf' });
```
devonzara, line 56: May also be worth testing that the appropriate response is given when _no data_ is provided.

devonzara (approved):
Few minor things, but otherwise looks good to me!

devonzara (changes requested):
Sorry, took a closer look at `updateUserById` and realized there's a logical error that needs to be fixed. Edited my previous comment on it.

devonzara (approved):
Changes look good. 👍🏼

Caleb-Cohen (approved):
LGTM!

PR #64, opened 2023-03-25
#33 Remove endpoint for getting a user's profile

### LINKED ISSUE #33: Remove endpoint for getting a user's profile
**Related Issues**
Project: #19 
Depends on: #9 

**Description**
The existing endpoint isn't very RESTful and it'll probably need to be replaced with the "get all users" functionality introduced in #17 .  So once #9 is completed, we'll use that for getting a user's profile.

Right now, we make a call to that endpoint to determine whether they're logged in [here](https://github.com/timmyichen/dev-directory/blob/760bb5e233473482d4f80f7bba92e37d328bd57b/pages/index.tsx#L56).  Once we remove the current route, we'll need to update this to use the route created in #9, which will require extracting the user's ID from the jwt auth token stored as a cookie.

The task is:
* Remove the existing `/api/users/` API endpoint
* Update the endpoint called in `index.tsx` to the new endpoint created in #9, using the cookie jwt auth token to determine the currently-logged-in-user's ID.

**Acceptance Criteria**
* Existing `/api/users/` endpoint is removed
* The app can still detect whether someone is logged in or not

windyScripts:
Closes #33 

# Description

- Updated API call end-point for get user
- Removed redundant route
- Added user id extraction using JWT as part of the get user call.
- Added universal-cookie as a dependency

## Type of change

# Checklist:

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,9 +1,12 @@
 import { Box, Button, Container, Typography } from '@mui/material';
 import { NextPage } from 'next';
 import React from 'react';
+import jwt from 'jsonwebtoken';
+import Cookies from 'universal-cookie';
 
 import createAxiosInstance from 'client/lib/axios';
 import { getDiscordOauthUrl } from 'client/lib/oauth';
+import { AUTH_COOKIE_NAME } from 'shared/constants';
 
 interface Props {
   isAuthed: boolean;
@@ -43,8 +46,12 @@ const Index: NextPage<Props> = (props) => {
 
 Index.getInitialProps = async ({ req }) => {
   try {
+    const cookies = new Cookies(req?.headers.cookie);
+    const token = cookies.get(AUTH_COOKIE_NAME);
+    const payload = jwt.decode(token) as jwt.JwtPayload;
+
     const axios = createAxiosInstance(req);
-    await axios.get('/api/users/');
+    await axios.get('/api/users/' + payload.user_id);
 
     return { isAuthed: true };
   } catch (error) {
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,7 +1,7 @@
 import { Box, Button, Container, Typography } from '@mui/material';
+import jwt from 'jsonwebtoken';
 import { NextPage } from 'next';
 import React from 'react';
-import jwt from 'jsonwebtoken';
 import Cookies from 'universal-cookie';
 
 import createAxiosInstance from 'client/lib/axios';
--- a/server/api/auth/auth.test.ts
+++ b/server/api/auth/auth.test.ts
@@ -63,7 +63,7 @@ describe('auth router', () => {
 
       const cookiePayload = jwt.decode(cookie.value);
 
-      const createdUser = await User.findOne({ where: { discord_user_id: id } });
+      const createdUser = await User.findOne({ where: { discord_user_id: id }});
 
       // the cookie payload will contain extra things, so we just
       // want to make sure that it contains the same fields as user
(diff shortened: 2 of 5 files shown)


timmyichen (approved):
Thanks for taking this on! I realized much too late that we should also be removing expired tokens, but that's outside the scope of this ticket - this will be enough to unblock the Pagination project! I'll add a ticket after this one is merged to invalidate expired tokens.

pages/index.tsx
```diff
@@ -43,8 +46,12 @@ const Index: NextPage<Props> = (props) => {
 
 Index.getInitialProps = async ({ req }) => {
   try {
+    const cookies = new Cookies(req?.headers.cookie);
+    const token = cookies.get(AUTH_COOKIE_NAME);
+    const payload = jwt.decode(token) as jwt.JwtPayload;
+
     const axios = createAxiosInstance(req);
-    await axios.get('/api/users/');
+    await axios.get('/api/users/' + payload.user_id);
```
Caleb-Cohen, line 54: How does this work?

In my local, my JWT is: `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoxLCJpYXQiOjE2Nzk3NzE0NjUsImV4cCI6MTY4MDM3NjI2NX0.aXXWWhweVc2R_mC8KtY9hqIGP8HlQQN4Ug5vv9OhqP4`

Which is decoded to

```JS

{
  "user_id": 1,
  "iat": 1679771465,
  "exp": 1680376265
}

```
How does whatever is issuing the JWT know my user_id?

Caleb-Cohen (approved):
Left a question, but LGTM!

pages/index.tsx
devonzara: The token is created when you login and it is sent to you as a cookie. Subsequent requests will include that cookie, thus letting the server know who you are.

During login (`/api/auth/login`), a `User` is found or created from your Discord user id:
https://github.com/timmyichen/dev-directory/blob/main/server/api/auth/auth.controller.ts#L37
https://github.com/timmyichen/dev-directory/blob/main/server/lib/auth.ts#L51-L59

That `User` (which includes your local user `id`) is then passed to a function that generates the JWT token:
https://github.com/timmyichen/dev-directory/blob/main/server/api/auth/auth.controller.ts#L38
https://github.com/timmyichen/dev-directory/blob/main/server/lib/auth.ts#L14-L18

Finally, that JWT token is then set as a cookie in the response:
https://github.com/timmyichen/dev-directory/blob/main/server/api/auth/auth.controller.ts#L40

pages/index.tsx
```diff
@@ -43,8 +46,12 @@ const Index: NextPage<Props> = (props) => {
 
 Index.getInitialProps = async ({ req }) => {
   try {
+    const cookies = new Cookies(req?.headers.cookie);
```
devonzara, line 49: What's the benefit of using `universal-cookie` here instead of `req?.cookies[AUTH_COOKIE_NAME]`?

Doing a quick search, I don't see `universal-cookie` being used anywhere else... could we possibly remove it as a dependency entirely? @timmyichen

server/api/auth/auth.test.ts
devonzara, line 1: nit: To keep commits small and file changes relevant to the issue, it's probably best to not include unrelated style fixes from other files.

server/lib/db.ts
devonzara, line 1: nit: Same as above, unrelated style fixes

pages/index.tsx
timmyichen: Needs to work on the client too!

devonzara: Fair, I suppose until the CookieStore API becomes stable and common in most browsers, a package is probably the easiest solution. 👍🏼

PR #65, opened 2023-03-26
Api all users

### LINKED ISSUE #57: Create new API route that returns all users
**Related Issues**
- Project: #17 
- Depends on: #33

**Description**
- Create new API route that returns all users: `GET /api/users`
- Return the same fields that are being returned by the `GET /api/users/:id` route.

**Acceptance Criteria**
- The route in question exists.
- Acceptance tests exist for this route.

Caleb-Cohen:
**Do not merge until #33 is completed**

Closes #57 (replace with issue number)

# Description

Created endpoint that returns all users.

Abstracted user test

## Testing

execute `npm run test`

## Type of change

- Change to documentation/GitHub flows/CICD

## Screenshots

Please include any screenshots if there are visual changes.

# Checklist:

### DIFF
--- a/server/api/users/users.controller.ts
+++ b/server/api/users/users.controller.ts
@@ -74,4 +74,25 @@ export const updateUserById: RequestHandler<{ id: string }, string, UpdatableFie
   }
 
   res.sendStatus(200);
+}
+
+export const getUsers: RequestHandler<UserProfile[]> = async (req, res) => {
+  const users = await User.findAll({
+    attributes: [
+      'id',
+      'discord_user_id',
+      'discord_name',
+      'bio',
+      'twitter_username',
+      'linkedin_url',
+      'github_username',
+      'website',
+    ],
+  });
+
+  if (users.length === 0) {
+    throw new NotFoundError('No Users found');
+  }
+
+  res.json(users);
 };
--- a/server/api/users/users.controller.ts
+++ b/server/api/users/users.controller.ts
@@ -19,15 +19,19 @@ export const getCurrentUser: RequestHandler<void, ClientUser> = (req, res) => {
 };
 
 export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req, res) => {
-  const allowedFields = User.allowedFields;
-  const user = await User.findByPk(req.params.id)
+  const user = await User.findByPk(req.params.id, { attributes: User.allowedFields });
 
   if (!user) {
     throw new NotFoundError('User not found');
   }
-  
-  const pickedUser = _.pick(user, allowedFields) as UserProfile
-  res.json(pickedUser);
+
+  res.json(user);
+};
+
+export const getUsers: RequestHandler<UserProfile[]> = async (req, res) => {
+  const users = await User.findAll({ attributes: User.allowedFields });
+
+  res.json(users);
 };
 
 interface UpdatableFields {
@@ -65,14 +69,4 @@ export const updateUserById: RequestHandler<{ id: string }, string, UpdatableFie
   }
 
   res.sendStatus(200);
-}
-
-export const getUsers: RequestHandler<UserProfile[]> = async (req, res) => {
-  const allowedFields = User.allowedFields;
-  const users = await User.findAll()
-
-
-
-  const pickedUser = _.map(users, user => _.pick(user, allowedFields));
-  res.json(pickedUser);
 };
--- a/server/api/users/users.controller.ts
+++ b/server/api/users/users.controller.ts
@@ -18,25 +18,16 @@ export const getCurrentUser: RequestHandler<void, ClientUser> = (req, res) => {
   res.json(filteredUser);
 };
 
-export const getUserById: RequestHandler<{ id: string }, UserProfile> = async (req, res) => {
-  const user = await User.findByPk(req.params.id, {
-    attributes: [
-      'id',
-      'discord_user_id',
-      'discord_name',
-      'bio',
-      'twitter_username',
-      'linkedin_url',
-      'github_username',
-      'website',
-    ],
-  });
+export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req, res) => {
+  const allowedFields = User.allowedFields;
+  const user = await User.findByPk(req.params.id)
 
   if (!user) {
     throw new NotFoundError('User not found');
   }
-
-  res.json(user);
+  
+  const pickedUser = _.pick(user, allowedFields) as UserProfile
+  res.json(pickedUser);
 };
 
 interface UpdatableFields {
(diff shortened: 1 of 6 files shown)


server/api/users/users.test.ts
```diff
@@ -24,16 +26,27 @@ describe('auth router', () => {
...
-        discord_name: user.discord_name,
-        bio: user.bio,
-        twitter_username: user.twitter_username,
-        linkedin_url: user.linkedin_url,
-        github_username: user.github_username,
-        website: user.website,
-      });
+      expect(res.body).toEqual(getExpectedUserObject(user));
+      // Delete the user that was created
+      await User.destroy({ where: {}});
```
Caleb-Cohen, line 31: Needed to delete User DB here because it was carrying over into the next test

server/test/utils.ts
```diff
@@ -66,4 +67,20 @@ async function createUser({
   return user;
 }
 
-export default createUser;
+function getExpectedUserObject(user: User) {
```
Caleb-Cohen, line 70: abstracted user compares for testing

Caleb-Cohen (commented):
added some comments

server/api/users/users.controller.ts
```diff
@@ -38,3 +38,24 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+      'id',
+      'discord_user_id',
+      'discord_name',
+      'bio',
+      'twitter_username',
+      'linkedin_url',
+      'github_username',
+      'website',
+    ],
+  });
```
timmyichen, line 54: So we do this pattern a lot it seems like - both here and down below we should just be using the lodash pick method. Can reuse the same array of allowed fields wherever we need it. Bonus - add a getter to the User model to extract these fields

server/api/users/users.controller.ts
```diff
@@ -38,3 +38,24 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+
+  if (users.length === 0) {
+    throw new NotFoundError('No Users found');
+  }
```
timmyichen, line 58: It's okay for there to be no users - the route can be described as "give me all the users", and "here are all the users, which is zero of them" is an okay response! I would not error here

server/api/users/users.test.ts
```diff
@@ -24,16 +26,27 @@ describe('auth router', () => {
...
+    });
+  });
+
+  describe('GET /', () => {
+    it('returns 404 if specified no users exist', async () => {
+      const res = await server.exec.get('/api/users/');
+      expect(res.status).toBe(404);
+    });
+
+    it('returns the specified user', async () => {
```
timmyichen, line 41: This should probably be renamed as it seems inaccurate for this new route

server/api/users/users.test.ts
timmyichen: I think it will be better to instead have the test suite spin up a new test app before every test, rather than before all tests. This solution may not work if, say, we need to test that route more, or if the User object gets some related models (because then we'd have to delete them all)

timmyichen (commented):
Mostly looks good, just left some comments about changes I think would be good

server/api/users/users.test.ts
```diff
@@ -24,16 +26,27 @@ describe('auth router', () => {
...
-        website: user.website,
-      });
+      expect(res.body).toEqual(getExpectedUserObject(user));
+      // Delete the user that was created
+      await User.destroy({ where: {}});
+    });
+  });
+
+  describe('GET /', () => {
+    it('returns 404 if specified no users exist', async () => {
```
devonzara, line 36: In relation to [tim's previous comment](https://github.com/timmyichen/dev-directory/pull/65/files#r1148667108), this test should probably expect to receive an empty array instead.

package.json
```diff
@@ -11,7 +11,7 @@
     "docker:db": "docker-compose exec pg psql -U postgres -d dev-directory",
     "docker:deps": "docker-compose up -d",
     "heroku-postbuild": "npm run build && npx sequelize-cli db:migrate",
-    "jest": "cross-env NODE_ENV=test npx jest --runInBand",
+    "jest": "cross-env NODE_ENV=test npx jest --runInBand --forceExit",
```
Caleb-Cohen, line 14: I could not get this to pass github actions testing without it. It would hang otherwise.

server/api/users/users.controller.ts
```diff
@@ -19,43 +19,23 @@ export const getCurrentUser: RequestHandler<void, ClientUser>  = (req, res) => {
...
-      'discord_name',
-      'bio',
-      'twitter_username',
-      'linkedin_url',
-      'github_username',
-      'website',
-    ],
-  });
+  const allowedFields = User.allowedFields;
+  const user = await User.findByPk(req.params.id)
```
devonzara, line 23: I'd probably inline `allowedFields` in the query; you can then delete lines 22 and 29 then return `user` in the response.
```js
const user = await User.findByPk(req.params.id, { attributes: User.allowedFields });
```

server/api/users/users.controller.ts
```diff
@@ -19,43 +19,23 @@ export const getCurrentUser: RequestHandler<void, ClientUser>  = (req, res) => {
...
-      'github_username',
-      'website',
-    ],
-  });
-
-  if (users.length === 0) {
-    throw new NotFoundError('No Users found');
-  }
+  const allowedFields = User.allowedFields;
+  const users = await User.findAll()
```
devonzara, line 35: Same as above... inline `allowedFields`. You can then delete lines 34, 39, and 3; then return `users` in the response.
```js
const users = await User.findAll({ attributes: allowedFields });
```

server/api/users/users.controller.ts
```diff
@@ -19,43 +19,23 @@ export const getCurrentUser: RequestHandler<void, ClientUser>  = (req, res) => {
...
+
+
 
-  res.json(users);
+  const pickedUser = _.map(users, user => _.pick(user, allowedFields));
```
devonzara, line 39: `pickedUser` will probably be deleted, but if not... it should be plural.

server/api/users/users.test.ts
```diff
@@ -28,17 +28,17 @@ describe('auth router', () => {
       expect(res.status).toBe(200);
       expect(res.body).toEqual(getExpectedUserObject(user));
       // Delete the user that was created
-      await User.destroy({ where: {}});
+      //await User.destroy({ where: {}});
```
devonzara: Remove comments if the `beforeEach` approach works.

devonzara (changes requested):
Looks good for the most part, few minor changes I'd suggest...

server/api/users/users.test.ts
```diff
@@ -24,16 +26,25 @@ describe('auth router', () => {
...
+    it('expects empty array if no users exist', async () => {
+      const res = await server.exec.get('/api/users/');
+      expect(res.body).toStrictEqual([]);
+    });
+
+    it('returns all users', async () => {
+      const userOne = await createUser();
+      const userTwo = await createUser();
+
+
```
devonzara, line 43: nit: New linting rules from #68 should prevent double empty lines, pull in the latest changes from `main` and re-run the linter.

server/api/users/users.controller.ts
Caleb-Cohen: Doesn't this do away with _.pick? Won't @timmyichen smite me for going against his request?

devonzara: Since you now have the field names extracted to `User.allowedFields`, there's no need to `_.pick()` when you can pass `User.allowedFields` to Sequalize and have it only select those fields from the get-go.

Thus eliminating the need for extra variables, looping over the results to cherry-pick items, and importing lodash at all.

MatthewBozin (approved):
LGTM

devonzara (approved):

Issue #39, opened 2023-03-15
Create seed script that inserts users into the database


timmyichen:
**Related Issues**
Depends on: #38

**Description**
Since dev relies on discord auth, it would be nice if there were a bunch of users already included in the database since we can't realistically manually create a bunch of users.

Using the `createUser` test utility function, let's create a file `server/scripts/seed.ts` that creates 100 random users and inserts them into the database.  This should be done within a [transaction](https://sequelize.org/docs/v6/other-topics/transactions/#managed-transactions) so that if it fails, it doesn't insert any data.

**Acceptance Criteria**
* `npm run seed` exists as a script, and running it inserts 100 random users into the database.

jmoldyvan:
ill try this one!

timmyichen:
> ill try this one!

Still depends on #38 which is currently in review!

jmoldyvan:
> > ill try this one!
> 
> Still depends on #38 which is currently in review!

I'll wait until it gets completed

MainlyColors:
I just merged my PR #48 that closes ticket #38 so your good to start this 🚀🚀🚀

jmoldyvan:
Sounds good I'll start working on it

Caleb-Cohen:
Hey @jmoldyvan, where are we at with this? I could really use it for #58.

jmoldyvan:
> Hey @jmoldyvan, where are we at with this? I could really use it for #58.

Just got my PR in, had some trouble and then some internet trouble.
I also updated the utils createUser function so that a bulkCreate would be possible, which may or may not be useful for #58

PR #73, opened 2023-03-31
Modify  /api/users to support pagination

### LINKED ISSUE #58: Modify `/api/users` API route to support pagination
**Related Issues**
- Project: #17 
- Depends on: #57  

**Description**
- Modify `/api/users` API route to support pagination using the Offset API Pagination
- This route needs to support a `page` query parameter.
- Limit the API route to return 20 profiles.
- Return which page we're on and the total pages available in the database.

[Read more about Offset API Pagination](https://nordicapis.com/understanding-5-types-of-web-api-pagination/#offset-api-pagination)

**Acceptance Criteria**
- The `/api/users` route can be paginated, returning 20 users at a time.
- Acceptance tests exist.

Caleb-Cohen:
Closes #58 

# Description

Modifies `/api/users` to support pagination

## Testing

A bit hard to test manual in local because #39 is not yet completed.

hitting `localhost:3000/api/users` should return first 20 users in db
hitting `localhost:3000/api/users?page=-1` or `localhost:3000/api/users?page=a` should return 400 Invalid page
hitting a page beyond homepage that has no users (I.E `localhost:3000/api/users?page=9001`) should return 400 Page out of bounds

can also execute `npm run test`

## Type of change

## Screenshots

No

# Checklist:

### DIFF
--- a/server/api/users/users.controller.ts
+++ b/server/api/users/users.controller.ts
@@ -28,33 +28,28 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
   res.json(user);
 };
 
-export const getUsers: RequestHandler<UserProfile[]> = async (req, res) => {
+export const getUsers: RequestHandler = async (req, res) => {
   const page = req.query.page ? parseInt(req.query.page as string, 10) : 1;
   const limit = 20;
   const offset = (page - 1) * limit;
 
   if (!Number.isInteger(page) || page < 1) {
-    return res.status(400).json({ error: 'Invalid page number' });
+    throw new BadRequestError('Invalid page number');
   }
 
-  // Fine all users, using the limit and offset to implement offset pagination
-  const users = await User.findAll({
+  const { rows: users, count } = await User.findAndCountAll({
     attributes: User.allowedFields,
     limit: limit,
     offset: offset,
   });
 
-  const count = await User.count();
   const totalPages = Math.ceil(count / limit);
 
   if (offset >= count && page !== 1) {
-    return res.status(400).json({ error: 'Page out of range' });
+    throw new BadRequestError('Page out of range');
   }
 
-  res.json({
-    page: page,
-    totalPages: totalPages,
-    users });
+  res.json({ page: page, totalPages: totalPages, users });
 };
 
 interface UpdatableFields {
--- a/server/api/users/users.controller.ts
+++ b/server/api/users/users.controller.ts
@@ -7,6 +7,8 @@ import { UserProfile } from 'server/types/User';
 
 type ClientUser = Pick<User, 'id' | 'discord_user_id'>
 
+export const USERS_LIMIT = 20;
+
 export const getCurrentUser: RequestHandler<void, ClientUser> = (req, res) => {
   // eslint-disable-next-line  @typescript-eslint/no-non-null-assertion
   const filteredUser = _.pick(req.user!, [
@@ -30,8 +32,7 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
 
 export const getUsers: RequestHandler = async (req, res) => {
   const page = req.query.page ? parseInt(req.query.page as string, 10) : 1;
-  const limit = 20;
-  const offset = (page - 1) * limit;
+  const offset = (page - 1) * USERS_LIMIT;
 
   if (!Number.isInteger(page) || page < 1) {
     throw new BadRequestError('Invalid page number');
@@ -39,14 +40,14 @@ export const getUsers: RequestHandler = async (req, res) => {
 
   const { rows: users, count } = await User.findAndCountAll({
     attributes: User.allowedFields,
-    limit: limit,
+    limit: USERS_LIMIT,
     offset: offset,
     order: [
       ['id', 'ASC'],
     ],
   });
 
-  const totalPages = Math.ceil(count / limit) || 1;
+  const totalPages = Math.ceil(count / USERS_LIMIT) || 1;
 
   if (offset >= count && page !== 1) {
     throw new BadRequestError('Page out of range');
--- a/server/api/users/users.controller.ts
+++ b/server/api/users/users.controller.ts
@@ -41,9 +41,12 @@ export const getUsers: RequestHandler = async (req, res) => {
     attributes: User.allowedFields,
     limit: limit,
     offset: offset,
+    order: [
+      ['id', 'ASC'],
+    ],
   });
 
-  const totalPages = Math.ceil(count / limit);
+  const totalPages = Math.ceil(count / limit) || 1;
 
   if (offset >= count && page !== 1) {
     throw new BadRequestError('Page out of range');
--- a/server/api/users/users.controller.ts
+++ b/server/api/users/users.controller.ts
@@ -29,9 +29,32 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
 };
 
 export const getUsers: RequestHandler<UserProfile[]> = async (req, res) => {
-  const users = await User.findAll({ attributes: User.allowedFields });
+  const page = req.query.page ? parseInt(req.query.page as string, 10) : 1;
+  const limit = 20;
+  const offset = (page - 1) * limit;
 
-  res.json(users);
+  if (!Number.isInteger(page) || page < 1) {
+    return res.status(400).json({ error: 'Invalid page number' });
+  }
+
+  // Fine all users, using the limit and offset to implement offset pagination
+  const users = await User.findAll({
+    attributes: User.allowedFields,
+    limit: limit,
+    offset: offset,
+  });
+
+  const count = await User.count();
+  const totalPages = Math.ceil(count / limit);
+
+  if (offset >= count && page !== 1) {
+    return res.status(400).json({ error: 'Page out of range' });
+  }
+
+  res.json({
+    page: page,
+    totalPages: totalPages,
+    users });
 };
 
 interface UpdatableFields {
--- a/server/api/users/users.test.ts
+++ b/server/api/users/users.test.ts
@@ -37,56 +37,54 @@ describe('auth router', () => {
   });
 
   describe('GET /', () => {
-    it('expects empty array if no users exist on the homepage', async () => {
+    it('API endpoint exists', async () => {
       const res = await server.exec.get('/api/users/');
+      expect(res.status).toBe(200);
+    });
+
+    it('expects empty array if no users exist', async () => {
+      const res = await server.exec.get('/api/users/');
+      expect(res.body.page).toEqual(1);
+      expect(res.body.totalPages).toEqual(0);
       expect(res.body.users).toStrictEqual([]);
     });
 
     it('expect invalid page request to return invalid page', async () => {
       const res = await server.exec.get('/api/users/?page=-1');
-      expect(res.status).toBe(400);
-      expect(res.body.error).toEqual('Invalid page number');
+      expect(res.body.message).toEqual('Invalid page number');
+    });
+
+    it('expect invalid page request to return invalid page', async () => {
+      const res = await server.exec.get('/api/users/?page=a');
+      expect(res.body.message).toEqual('Invalid page number');
     });
 
     it('expect page request beyond page limit to be out of range', async () => {
       const res = await server.exec.get('/api/users/?page=100');
-      expect(res.status).toBe(400);
-      expect(res.body.error).toEqual('Page out of range');
+      expect(res.body.message).toEqual('Page out of range');
     });
 
-    it('returns first 20 users when it hits the homepage', async () => {
-
-      const users: User[] = [];
-
-      for(let i = 0; i < 25; i++){
-        const user = await createUser();
-        users.push(user);
-      }
+    it('returns an array of 20 users', async () => {
+      const count = 25;
+      const users: User[] = await Promise.all(Array.from({ length: count }, createUser));
+      users.sort((a, b) => a.id - b.id);
 
       const limit = 20;
-      const count = await User.count();
       const totalPages = Math.ceil(count / limit);
 
       const res = await server.exec.get('/api/users/');
       expect(res.status).toBe(200);
       expect(res.body.users.length).toEqual(limit);
-      expect(res.body.users[0]).toEqual(getExpectedUserObject(users[0]));
-      expect(res.body.users[19]).toEqual(getExpectedUserObject(users[19]));
       expect(res.body.page).toEqual(1);
       expect(res.body.totalPages).toEqual(totalPages);
     });
 
     it('returns last 5 users when it hits the second page', async () => {
-
-      const users: User[] = [];
-
-      for(let i = 0; i < 25; i++){
-        const user = await createUser();
-        users.push(user);
-      }
+      const count = 25;
+      const users: User[] = await Promise.all(Array.from({ length: count }, createUser));
+      users.sort((a, b) => a.id - b.id);
 
       const limit = 20;
-      const count = await User.count();
       const totalPages = Math.ceil(count / limit);
 
       const res = await server.exec.get('/api/users?page=2');
@@ -94,7 +92,6 @@ describe('auth router', () => {
       expect(res.status).toBe(200);
       expect(res.body.users.length).toEqual(5);
       expect(res.body.users[0]).toEqual(getExpectedUserObject(users[20]));
-      expect(res.body.users[4]).toEqual(getExpectedUserObject(users[24]));
       expect(res.body.page).toEqual(2);
       expect(res.body.totalPages).toEqual(totalPages);
     });
--- a/server/api/users/users.test.ts
+++ b/server/api/users/users.test.ts
@@ -3,7 +3,9 @@ import TestServer from 'server/test/server';
 import { createUser, getExpectedUserObject } from 'server/test/utils';
 import { UserProfile } from 'server/types/User';
 
-describe('auth router', () => {
+import { USERS_LIMIT } from './users.controller';
+
+describe('user router', () => {
   let server: TestServer;
 
   beforeAll(async () => {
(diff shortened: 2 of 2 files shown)


server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+      expect(res.body.error).toEqual('Page out of range');
+    });
+
+    it('returns first 20 users when it hits the homepage', async () => {
+
+      const users: User[] = [];
+
+      for(let i = 0; i < 25; i++){
+        const user = await createUser();
+        users.push(user);
```
Caleb-Cohen, line 63: I could not develop a way to have users stay between this test and the test that starts on line 79. It looks like @timmyichen is good at foreshadowing potential issues with the afterEach deleting users.

Open to suggestions, or we can refactor after we solve the testing/db issue.

server/api/users/users.controller.ts
```diff
@@ -29,9 +29,32 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+  const users = await User.findAll({
+    attributes: User.allowedFields,
+    limit: limit,
+    offset: offset,
+  });
+
+  const count = await User.count();
+  const totalPages = Math.ceil(count / limit);
+
+  if (offset >= count && page !== 1) {
```
Caleb-Cohen, line 50: I wanted to exclude 1 so that homepage would not report Page out of range if the DB is empty

MainlyColors (approved):
LGTM

Tests pass on local via `npm run test:server` & with manual testing on dev server after inserting some users manually 
image

server/api/users/users.controller.ts
```diff
@@ -29,9 +29,32 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
 };
 
 export const getUsers: RequestHandler<UserProfile[]> = async (req, res) => {
```
timmyichen, line 31: i just realized this line - what is this `UserProfile[]` trying to type?  i believe the first argument for `RequestHandler` is the params.  if you're trying to type the request response, you'll have to also include the pagination info

server/api/users/users.controller.ts
```diff
@@ -29,9 +29,32 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+  }
+
+  // Fine all users, using the limit and offset to implement offset pagination
+  const users = await User.findAll({
+    attributes: User.allowedFields,
+    limit: limit,
+    offset: offset,
+  });
+
+  const count = await User.count();
```
timmyichen, line 47: these two can (and probably should) happen concurrently via `Promise.all`

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
   });
 
   describe('GET /', () => {
-    it('expects empty array if no users exist', async () => {
+    it('expects empty array if no users exist on the homepage', async () => {
```
timmyichen, line 40: API routes don't really have a relation to pages - a homepage may call an API route but an API route is often crafted independently of the page it will used on.  i think we can just remove this added string

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+      expect(res.body.users).toStrictEqual([]);
     });
 
-    it('returns all users', async () => {
-      const userOne = await createUser();
-      const userTwo = await createUser();
+    it('expect invalid page request to return invalid page', async () => {
+      const res = await server.exec.get('/api/users/?page=-1');
+      expect(res.status).toBe(400);
+      expect(res.body.error).toEqual('Invalid page number');
```
timmyichen, line 48: when using `BadRequestError`, you'll need to use `res.body.message` here

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
   });
 
   describe('GET /', () => {
-    it('expects empty array if no users exist', async () => {
+    it('expects empty array if no users exist on the homepage', async () => {
       const res = await server.exec.get('/api/users/');
-      expect(res.body).toStrictEqual([]);
+      expect(res.body.users).toStrictEqual([]);
```
timmyichen, line 42: we should also validate the entire pagination object values here

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+    });
+
+    it('expect page request beyond page limit to be out of range', async () => {
+      const res = await server.exec.get('/api/users/?page=100');
+      expect(res.status).toBe(400);
+      expect(res.body.error).toEqual('Page out of range');
+    });
+
+    it('returns first 20 users when it hits the homepage', async () => {
+
```
timmyichen, line 58: nit: remove empty line

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+      expect(res.body.error).toEqual('Invalid page number');
+    });
+
+    it('expect page request beyond page limit to be out of range', async () => {
+      const res = await server.exec.get('/api/users/?page=100');
+      expect(res.status).toBe(400);
+      expect(res.body.error).toEqual('Page out of range');
+    });
+
+    it('returns first 20 users when it hits the homepage', async () => {
```
timmyichen, line 57: similar comment about mentioning the homepage

server/api/users/users.test.ts
timmyichen, line 63: we can parallelize this a bit by using `Promise.all` and a map over an empty array

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+
+      const users: User[] = [];
+
+      for(let i = 0; i < 25; i++){
+        const user = await createUser();
+        users.push(user);
+      }
+
+      const limit = 20;
+      const count = await User.count();
```
timmyichen, line 67: you can just use the number used in the for loop; you should know how many users exist in the DB at this point

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+      const totalPages = Math.ceil(count / limit);
 
       const res = await server.exec.get('/api/users/');
       expect(res.status).toBe(200);
-      expect(res.body[0]).toEqual(getExpectedUserObject(userOne));
-      expect(res.body[1]).toEqual(getExpectedUserObject(userTwo));
+      expect(res.body.users.length).toEqual(limit);
+      expect(res.body.users[0]).toEqual(getExpectedUserObject(users[0]));
+      expect(res.body.users[19]).toEqual(getExpectedUserObject(users[19]));
```
timmyichen, line 74: i dont think we actually need to validate that the first and last users match, only we are receiving 20 users.  20 distinct users if you want to go the extra mile but i don't think it's necessary

server/api/users/users.test.ts
timmyichen: i think it's fine to just do what you're doing (just parallelize it a bit), as long as you're not seeing a big perf hit

timmyichen (commented):
great work! let me know if you have any questions about any of the comments

timmyichen (commented):
great work! let me know if you have any questions about any of the comments

server/api/users/users.controller.ts
```diff
@@ -29,9 +29,32 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
 };
 
 export const getUsers: RequestHandler<UserProfile[]> = async (req, res) => {
-  const users = await User.findAll({ attributes: User.allowedFields });
+  const page = req.query.page ? parseInt(req.query.page as string, 10) : 1;
```
devonzara, line 32: This may be open for debate, but I'd argue it should fallback to 1 if an invalid value is provided. If that's the case, we could rewrite this as:
```suggestion
  const page = parseInt(req.query.page as string, 10) || 1;
```

server/api/users/users.controller.ts
```diff
@@ -29,9 +29,32 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+  const limit = 20;
+  const offset = (page - 1) * limit;
 
-  res.json(users);
+  if (!Number.isInteger(page) || page < 1) {
+    return res.status(400).json({ error: 'Invalid page number' });
+  }
```
devonzara, line 38: You've already made sure `page` will be an integer, therefore you wouldn't need `!Number.isInteger(page)`.

```suggestion
  if (page < 1) {
    return res.status(400).json({ error: 'Invalid page number' });
  }
```

Additionally, if we default to `1` above when invalid input is provided, we can eliminate this block entirely.
```suggestion
```

server/api/users/users.controller.ts
devonzara, line 47: This can be simplified using `findAndCountAll()`

server/api/users/users.controller.ts
```diff
@@ -29,9 +29,32 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+    limit: limit,
+    offset: offset,
+  });
+
+  const count = await User.count();
+  const totalPages = Math.ceil(count / limit);
+
+  if (offset >= count && page !== 1) {
+    return res.status(400).json({ error: 'Page out of range' });
+  }
```
devonzara, line 52: I'd probably remove this check and just return an empty array along with the standard metadata... the client side can conditionally render based on the length of the returned `users` array.

server/api/users/users.controller.ts
```diff
@@ -29,9 +29,32 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
...
+
+  res.json({
+    page: page,
+    totalPages: totalPages,
+    users });
```
devonzara, line 57: nit:
```suggestion
  res.json({ page, totalPages, users });
```

server/api/users/users.test.ts
devonzara, line 58: The API doesn't know about the homepage
```suggestion
    it('returns an array of 20 users', async () => {
```

server/api/users/users.test.ts
devonzara: I wouldn't worry about it... you obviously don't want to have to seed a lot of stuff into the database on each test, but it won't hurt to do it a few times.

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+      expect(res.body.page).toEqual(1);
+      expect(res.body.totalPages).toEqual(totalPages);
```
devonzara, line 76: This is testing a lot of stuff and it's being re-tested below. It should probably be broken up into smaller tests and the one below can then be simplified too.

I'd probably break it out into these tests:
- API endpoint exists
  - Simple test, hit the endpoint, check for status `200`. No need to create users.
- Number of items returned (this test's main goal)
  - Create 25 users, hit endpoint, check length of `res.body.users`.
- Sort order of items
  - Create 5 users, hit endpoint, `expect(_.map(res.body.users)).toEqual(_.map(users, 'id'), 'id')`
- Validity of metadata returned
  - Break into multiple tests to check 0 users, 20 users, 21 users
  - In the future, when we add customizable limits, we can reduce the counts for better test performance

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+    });
+
+    it('returns last 5 users when it hits the second page', async () => {
+
+      const users: User[] = [];
+
+      for(let i = 0; i < 25; i++){
+        const user = await createUser();
+        users.push(user);
+      }
```
devonzara, line 86: This can be optimized to:
```suggestion
      const users: User[] = await Promise.all(Array.from({ length: 25 }, createUser));
```

server/api/users/users.test.ts
```diff
@@ -37,19 +37,66 @@ describe('auth router', () => {
...
+      const totalPages = Math.ceil(count / limit);
+
+      const res = await server.exec.get('/api/users?page=2');
+
+      expect(res.status).toBe(200);
+      expect(res.body.users.length).toEqual(5);
+      expect(res.body.users[0]).toEqual(getExpectedUserObject(users[20]));
+      expect(res.body.users[4]).toEqual(getExpectedUserObject(users[24]));
+      expect(res.body.page).toEqual(2);
+      expect(res.body.totalPages).toEqual(totalPages);
```
devonzara, line 99: Assuming the previous test was split up, this test now only needs to check:
 - `res.body.users.length` equals `5`
 - `_.map(res.body.users, 'id')` equals `_.map(users, 'id')`

devonzara (changes requested):
I see Tim submit his review before I got around to finishing mine... sorry if there's an overlap in comments. 😅

server/api/users/users.controller.ts
devonzara: I'll agree and I'll raise you a `findAndCountAll()` ([Docs](https://sequelize.org/docs/v6/core-concepts/model-querying-finders/#findandcountall))

server/api/users/users.controller.ts
timmyichen: i'll happily take the other side - if the user provides an invalid input, we should tell them so, otherwise we could be doing something wrong on the frontend but not realizing it because we are getting a 200 response from the server with actual data.

server/api/users/users.controller.ts
timmyichen: It may not be an integer - `parseInt` can return `NaN` if a non-integer-like value is passed.  e.g. running the following in your browser

```js
Number.isInteger(parseInt("asdf"))
```

server/api/users/users.controller.ts
timmyichen: i was thinking we could add a linting rule for this since it happens a few times throughout this PR haha

server/api/users/users.controller.ts
devonzara: I feel like most APIs I've encountered in the wild will just respond with the first page...

| Site | `page=1` or `start=0` | `page=2` or `start=1` | `page=foobar` or `start=foobar` |
| :--- | :--- | :--- | :--- |
| **Reques** | [Link](https://reqres.in/api/users?page=1) | [Link](https://reqres.in/api/users?page=2) | [Link](https://reqres.in/api/users?page=foobar) |
| **Unsplash** | [Link](https://unsplash.com/napi/search/photos?query=nerd&page=1) | [Link](https://unsplash.com/napi/search/photos?query=nerd&page=2) | [Link](https://unsplash.com/napi/search/photos?query=nerd&page=foobar) |
| **Newegg** | [Link](https://www.newegg.com/p/pl?d=floppy+disks&page=1) | [Link](https://www.newegg.com/p/pl?d=floppy+disks&page=2) | [Link](https://www.newegg.com/p/pl?d=floppy+disks&page=foobar) |
| **Unreal Engine Marketplace** | [Link](https://www.unrealengine.com/marketplace/api/assets?start=0) | [Link](https://www.unrealengine.com/marketplace/api/assets?start=1) | [Link](https://www.unrealengine.com/marketplace/api/assets?start=foobar) |

I tried to find random sites in different markets... notice how their invalid values still return the same results as their `page=0`/`start=1` while `page=1`/`start=2` returns the next set of results.

server/api/users/users.controller.ts
devonzara: Ahh, good catch. Wouldn't have to worry about it though if we defaulted to 1 when not given a valid integer. 😜

server/api/users/users.controller.ts
timmyichen: I would argue that all those APIs are designed poorly.  APIs should not fail silently.

An example of a bad contract:
Client: give me the results for page 0
Server: here are the results for page 1
Client: give me the results for page 100
Server here are the results for page 1
Client: give me the results for page 1
Server: here are the results for page 1

An example of a better contract:
Client: give me the results for page 0
Server: we don't have a page zero, pages start at 1
Client: give me the results for page 100
Server: we don't have a page 100
Client: give me the results for page 1
Server: here are the results for page 1

server/api/users/users.controller.ts
devonzara: @timmyichen likewise, though I wasn't sure which settings to use. `always`? `avoidQuotes`? 
[ESLint Docs: object-shorthand](https://eslint.org/docs/latest/rules/object-shorthand)

I already have a branch with some other rules on it, I can add this one to it.

timmyichen: Yup I think always + avoid Quotes is good!

server/api/users/users.controller.ts
devonzara: I mean, I don't disagree... but that doesn't appear to be convention unless I'm just looking at all the wrong sites.

[Even GitHub's API will default to page 1 when given an invalid page parameter](https://api.github.com/search/repositories?q=javascript&page=foobar), unless it's numerical and over their limit, then it returns an error. 🤦🏼‍♂️

The first site I found that doesn't do that is Amazon, but it doesn't error either... instead, it removes the `page` query param from the URL, thus [redirecting back to page 1](https://www.amazon.com/s?k=nerds&page=foobar). So, basically the same thing, just accomplished differently. 😂

Caleb-Cohen: I'm going to agree with Tim on this. I think that a user requesting an invalid page should return a distinct error, not redirect to homepage.

server/api/users/users.controller.ts
Caleb-Cohen: you nit and request to change linting has been DENIED

server/api/users/users.test.ts
Caleb-Cohen: Is "Sort order of items" the correct name here? Wouldn't something like "created users match expected output of users" be more accurate?

Caleb-Cohen: Why do we need the metadata response test?

I plan on checking if the API endpoint exists that the length of the array, page, and totalPages matches in the 20-user test.

I'll have the new commits shortly and it may make it more clear.

server/api/users/users.controller.ts
Caleb-Cohen: I don't believe I understand what you are requesting.

server/api/users/users.test.ts
Caleb-Cohen: I'll need some clarification on how to split these tests up properly, but I've made some changes in the newest commit.

server/api/users/users.test.ts
```diff
@@ -37,19 +39,63 @@ describe('auth router', () => {
...
+      users.sort((a,b) => a.id - b.id)
+
+      const limit = 20;
+      const totalPages = Math.ceil(count / limit);
+
+      const res = await server.exec.get('/api/users?page=2');
+
+      expect(res.status).toBe(200);
+      expect(res.body.users.length).toEqual(5);
+      expect(res.body.users[0]).toEqual(getExpectedUserObject(users[20]));
```
Caleb-Cohen, line 96: @timmyichen @devonzara 

This is occasionally failing. I believe it has to do with the Promise.all not guaranteeing ids are put into the array in sequential order. I've attempted a `users.sort((a,b) => a.id - b.id)` but that has not fixed the anomaly.

server/api/users/users.test.ts
Caleb-Cohen: Here's what I've done to satisfy your requests

1. API endpoint exists: I've created an API endpoint exists test
2. Number of items returned: I kept returns an array of 20 users, and returns last 5 users when it hits second page. I think checking the max length and second page is important using `res.body.users.length`
3. Sort order of items: was not created. I'm not entirely sure how to implement this request using the example you provided.
4. Validity of metadata in response: was not created. I have a comment above addressing it.

server/api/users/users.controller.ts
```diff
@@ -41,9 +41,12 @@ export const getUsers: RequestHandler = async (req, res) => {
     attributes: User.allowedFields,
     limit: limit,
     offset: offset,
+    order: [
+      ['id', 'ASC'],
+    ],
```
Caleb-Cohen, line 46: orders the user objects by Id since they are unordered with promise.all

server/api/users/users.controller.ts
```diff
@@ -41,9 +41,12 @@ export const getUsers: RequestHandler = async (req, res) => {
...
   });
 
-  const totalPages = Math.ceil(count / limit);
+  const totalPages = Math.ceil(count / limit) || 1;
```
Caleb-Cohen, line 49: empty database was returning 0 total pages

server/api/users/users.test.ts
```diff
@@ -36,21 +37,108 @@ describe('auth router', () => {
...
-    it('expects empty array if no users exist', async () => {
+  describe('GET /api/users', () => {
+    it('should have an API endpoint that exists', async () => {
       const res = await server.exec.get('/api/users/');
-      expect(res.body).toStrictEqual([]);
+
+      expect(res.status).toBe(200);
+    });
+
+    it('should return empty array if no users exist', async () => {
```
timmyichen, line 47: i think it would be nice to easily distinguish between happy path and unhappy path test cases by having unhappy cases start with "should error" and happy cases not.  So this one might instead be:

```
"should error with "empty array if no users exist"'
```

server/api/users/users.test.ts
```diff
@@ -36,21 +37,108 @@ describe('auth router', () => {
...
+      const res = await server.exec.get('/api/users/');
+
+      expect(res.body.users).toStrictEqual([]);
+    });
+
+    it('should return invalid page number with negative pages', async () => {
+      const res = await server.exec.get('/api/users/?page=-1');
+
+      expect(res.body.message).toEqual('Invalid page number');
+      expect(res.status).toBe(400);
```
timmyichen, line 57: nit: status checks should always be first imo (i.e. check status before checking more specific body values)

server/api/users/users.controller.ts
```diff
@@ -28,10 +28,31 @@ export const getUserById: RequestHandler<{id: string}, UserProfile> = async (req
   res.json(user);
 };
 
-export const getUsers: RequestHandler<UserProfile[]> = async (req, res) => {
-  const users = await User.findAll({ attributes: User.allowedFields });
+export const getUsers: RequestHandler = async (req, res) => {
+  const page = req.query.page ? parseInt(req.query.page as string, 10) : 1;
+  const limit = 20;
```
timmyichen, line 33: this might be useful as a constant.  i'd move this to the top of the file and capitalize it, e.g.

```js
const LIMIT = 20;
```

(capitalizing constants is a convention)

server/api/users/users.test.ts
```diff
@@ -1,6 +1,7 @@
 import { User } from 'server/models';
 import TestServer from 'server/test/server';
 import { createUser, getExpectedUserObject } from 'server/test/utils';
+import { UserProfile } from 'server/types/User';
 
 describe('auth router', () => {
```
timmyichen, line 6: lol whoops this is a copy paste error but this should be `users router` 🙃

server/api/users/users.test.ts
```diff
@@ -36,21 +37,108 @@ describe('auth router', () => {
...
+
+      const res = await server.exec.get('/api/users/');
+
+      expect(res.body.page).toEqual(1);
+      expect(res.body.totalPages).toEqual(2);
+    });
+
+    it('should return proper page metadata with the second page', async () => {
+      const count = 25;
+      await Promise.all(Array.from({ length: count }, createUser));
```
timmyichen, line 103: if `count` isn't being reused anywhere we can just use count inline:

```suggestion
      await Promise.all(Array.from({ length: 25 }, createUser));
```

server/api/users/users.controller.ts
timmyichen: ah yeah, seeing more in the tests - we should export this as a constant at the top level, then import it in the tests when checking for page size (i.e. if we change hte page size from 20 to 40 we wouldn't have to update all the tests too)

timmyichen (commented):
mostly LGTM, just some final changes!

server/api/users/users.test.ts
Caleb-Cohen: Why should it error when it's returning a 200 status code?

timmyichen: oops - meant to leave this comment on the next one that actually errors 🤡

devonzara (approved):

PR #74, opened 2023-03-31
#39 create seed script

### LINKED ISSUE #39: Create seed script that inserts users into the database

jmoldyvan:
Closes #39 

# Description

This PR creates a seed.ts file and the ability to run that file as the scrip `npm run seed script` which will insert 100 random users into the database via a transcript.
The PR also adds a makeUser function to the server/test/utils.ts file in order to use bulkCreate when creating the 100 users to reduce the amount of db inserts, improving efficiency. 

## Testing

Enter `npm run seed` and then check the db is 100 users is created

## Type of change

# Checklist:


tsconfig.json
```diff
@@ -41,7 +41,8 @@
     "server/**/*",
     "client/**/*",
     "shared/**/*",
-    "pages/**/*"
+    "pages/**/*",
+    "seed.ts"
```
timmyichen, line 45: this shouldn't be necessary as `seed.ts` lives in the server directory

server/types/User.ts
```diff
@@ -8,3 +8,14 @@ export type UserProfile = Pick<User, 'id'
...
+export interface UserObject {
+  email: string;
+  discord_user_id: string;
+  discord_name: string;
+  bio: string;
+  twitter_username: string;
+  linkedin_url: string;
+  github_username: string;
+  website: string;
+}
```
timmyichen, line 21: i think this is basically the `User` type without the `id` field; you can use `Omit` to perform the opposite of `Pick` to remove fields:

```typescript
export type UserFields = Omit<User, 'id'>;
```

server/scripts/seed.ts
```diff
@@ -0,0 +1,36 @@
+import Db from 'server/lib/db';
+import log from 'server/lib/log';
+import User from 'server/models/User.model';
+import { makeUser } from 'server/test/utils';
+import { UserObject } from 'server/types/User';
+
+async function create100Users(): Promise<UserObject[]> {
```
timmyichen, line 7: I think this would actually be pretty useful as a test utility script as "create many users", which allows you to pass an arbitrary number of users to create, then to just call that function here, passing 100.

timmyichen, line 7: I think this would actually be pretty useful as a test utility script as "create many users", which allows you to pass an arbitrary number of users to create, then to just call that function here, passing 100.

server/scripts/seed.ts
```diff
@@ -0,0 +1,36 @@
...
+  } catch (error) {
+    console.log(error);
+    log('Error inserting users');
+  }
+}
+
+export async function seed(): Promise<void> {
+  try {
+    const userArray = await create100Users();
+    await insertUsers(userArray);
```
timmyichen, line 28: any reason why these are split up?  is there any reason we would want to generate multiple user creation objects, but _not_ insert them?  if not, then i think it would be appropriate to just do the insertion in the same function that "creates" the users

server/test/utils.ts
```diff
@@ -60,11 +49,12 @@ async function createUser({
     github_username: github_username ?? randomEmptyChance(20, randUserName()),
     website: website ?? randomEmptyChance(20, randUrl()),
   };
+}
 
+async function createUser(options: Partial<UserObject> = {}) {
```
timmyichen, line 54: i think we need more distinctive names. what's the difference between `makeUser` and `createUser`?  "make" and "create" are synonyms so for someone unfamiliar with the codebase, it may be hard to understand when to use each function.

timmyichen (commented):
thanks for doing this! i left some comments around organization but the core of it looks good!

server/scripts/seed.ts
```diff
@@ -0,0 +1,36 @@
...
+import log from 'server/lib/log';
+import User from 'server/models/User.model';
+import { makeUser } from 'server/test/utils';
+import { UserObject } from 'server/types/User';
+
+async function create100Users(): Promise<UserObject[]> {
+  const promises = Array.from({ length: 100 }, makeUser);
+  const userArray = await Promise.all(promises);
+  return userArray;
+}
```
devonzara, line 11: 1. To keep consistent verb usage, I'd name this `make100Users()` since it only builds the objects, whereas `create` would be inserting them.

2. `makeUser` is not an `async` function and therefore shouldn't require being wrapped in `Promise.all()`.

3. I agree, this would be useful in testing if it were more dynamic, perhaps naming it just `makeUsers`. In which case, you could probably move it into the `Utils` file next to `makeUser()` and just import it.

```suggestion
function makeUsers(count = 100): UserObject[] {
  return Array.from({ length: count }, makeUser);
}
```

server/scripts/seed.ts
```diff
@@ -0,0 +1,36 @@
...
+
+async function insertUsers(userArray: UserObject[]) {
+  try {
+    await Db.sequelize.transaction(async transaction => {
+      await User.bulkCreate(userArray, { transaction });
+    });
+    log('Successfully inserted 100 users');
```
devonzara, line 18: Since there's no validation here, we should make the number dynamic.

```suggestion
    log(`Created ${userArray.length} users!`);
```

server/scripts/seed.ts
devonzara: It's essentially a 1-liner after a quick refactor, I'd throw it in `Utils`, honestly. Might be handy if we do anything with accessors or mutators where we want to have the raw data for the users prior to insertion.

server/test/utils.ts
devonzara: That one was on me, something I stole from Laravel. lol
[make](https://laravel.com/api/10.x/Illuminate/Database/Eloquent/Builder.html#method_make) makes an unsaved model while [create](https://laravel.com/api/10.x/Illuminate/Database/Eloquent/Builder.html#method_create) creates a saved model.

I agree, the names aren't very intuitive but for some reason I've never gotten them confused. 😂

server/types/User.ts
jmoldyvan: so would it be better for me to use the User type and omit? or leave as is?

timmyichen: I'd use omit, less maintenance and less redundancy

server/scripts/seed.ts
jmoldyvan: I'm not sure what you are referencing here?

server/test/utils.ts
jmoldyvan: yeah I also think changing may be beneficial?

server/scripts/seed.ts
jmoldyvan: ok yeah i can combine them

server/types/User.ts
jmoldyvan: i think I'm having a typescript issue, the new Omit export isnt working for these two lines of code

function createManyUsers(numberOfUsers: number): UserObject[] {
  return Array.from({ length: numberOfUsers }, makeUnsavedUserModel);
}

jmoldyvan:
So I'm pretty sure I did everything part of the comments left, however I couldn't get the 

export type UserFields = Omit<User, 'id'>;

to work with the updated createManyUsers function... I wanted to commit so I didn't lose any progress, but accidently pushed out of habit

server/types/User.ts
```diff
@@ -10,12 +10,12 @@ export type UserProfile = Pick<User, 'id'
...
-}
+    email: string;
+    discord_user_id: string;
+    discord_name: string;
+    bio: string;
+    twitter_username: string;
+    linkedin_url: string;
+    github_username: string;
+    website: string;
+  }
```
devonzara, line 21: nit: fix indentation

server/test/utils.ts
```diff
@@ -51,20 +51,25 @@ function makeUser({
...
   const pickedUser = _.pick(user, allowedFields);
   return pickedUser as User;
 }
 
 export {
-  makeUser,
-  createUser,
+  makeUnsavedUserModel,
+  createManyUsers,
+  createSavedUserModel,
```
devonzara, line 73: Careful when using the word Model as that implies we're returning an actual User Model. `makeUnsavedUserModel` is just returning a plain object with properties that match our User fields. It's not returning an actual Sequelize model (instance of `server/models/User.model.ts`).

Perhaps keeping `createUser` the same, as I think that's pretty straightforward... but then being more verbose on the others? eg. `makeUserObject`/`makeUserObjects` or `generateUserObject`/`generateUserObjects`?

server/scripts/seed.ts
```diff
@@ -1,36 +1,21 @@
...
-
-async function insertUsers(userArray: UserObject[]) {
+export async function seed(): Promise<void> {
   try {
+    const userArray = await createManyUsers(100);
     await Db.sequelize.transaction(async transaction => {
       await User.bulkCreate(userArray, { transaction });
     });
-    log('Successfully inserted 100 users');
+    log(`Created ${userArray.length} users!`);
```
devonzara: Line 12 doesn't need to be dynamic if the function itself isn't dynamic. Also, perhaps we should have moved `insertUsers` to Utils and renamed it to `createUsers(count)` (or similar) so it could be used in tests?

jmoldyvan: I'm not 100% sure what correction you are asking for...

server/test/utils.ts
jmoldyvan: alright, I'll change that up

server/scripts/seed.ts
jmoldyvan: i can make it dynamic by adding a parameter?

server/types/User.ts
jmoldyvan: nvm i got it!

jmoldyvan:
Finished up all the requested changes. Got the Omit working. Made the seed function itself dynamic?
But yeah, got it all good to go

server/scripts/seed.ts
devonzara: I'm not 100% sure what I was suggesting as I wasn't entirely certain what @timmyichen was suggesting in their review, so I was trying to leave it open for discussion. 😂

I think @timmyichen previously suggested merging the two functions, which is essentially what you did. I'm uncertain what direction he was going with that.

My initial suggestion was to move the function to Utils, give it a name that matches whatever pattern we decide on for the Utils functions (eg. `createUsers`), have it return an array of the inserted users, and export the function. Then `seed` can import and use that function to create its users. That would also allow us to also use `createUsers` (or w/e name it's given) in tests.

server/test/utils.ts
```diff
@@ -60,11 +49,16 @@ async function createUser({
...
 
+async function createUser(options: Partial<UserObject> = {}) {
+  const userObject = makeUserObject(options);
   // insert into DB after creation
-  const user = await User.create(userObject);
+  return await User.create(userObject);
+}
 
-  return user;
+function createManyUsers(numberOfUsers: number): Partial<UserObject>[] {
```
devonzara, line 60: Given the current naming, this `makeUserObjects` (plural form of `makeUserObject`) as it's not inserting the users.

server/scripts/seed.ts
devonzara: Re-reading his comments, I believe he was suggesting merging `insertUsers` into `create100Users` and then extracting that function out into Utils... so we agreed on the latter part, it should be made into a reusable function in Utils.

The difference in our suggestions was that I would not have merged them and just moved them both. Either works, since there aren't any current use cases for each, merging is fine. If, in the future, someone needs access to the raw data before the users are inserted, they can always extract it to a separate function themself.

jmoldyvan:
So I'm not sure why but it's not working on my local machine now...

server/test/utils.ts
```diff
@@ -74,6 +72,8 @@ function getExpectedUserObject(user: User) {
 }
 
 export {
+  makeUserObject,
+  createManyUsers,
```
devonzara, line 76: I'd vote for `createUsers` over `createManyUsers`, personally.

Aside from that, are we happy with the distinction between `makeUserObject` and `createUser`? If not, any other suggestions? @timmyichen

MainlyColors: How about `generateUserObject`

This was chatGPTs suggestion: `generateRandomUserObjectWithOptionalProperties`

Caleb-Cohen:
> So I'm not sure why but it's not working on my local machine now...

It looks like it's working on mine.

server/test/utils.ts
devonzara: @MainlyColors that was the other name [I was considering](https://github.com/timmyichen/dev-directory/pull/74#discussion_r1155095468) as well. 

ChatGPT clearly doesn't understand the idea of _human_ maintainable code. 😂

devonzara:
@Caleb-Cohen I believe his comment was before our pair programming session yesterday evening, there's been a commit since then and all CI tests have passed. 👍🏼

server/test/utils.ts
jmoldyvan: sure i can change the name

server/test/utils.ts
```diff
@@ -50,7 +40,7 @@ async function createUser({
   // build object
```
timmyichen, line 40: on the line above this, we should just be using `UserObject` instead of `Partial<UserObject>` since this function is expected to return values identical to the types we're expecting in the user model

server/test/utils.ts
```diff
@@ -60,11 +50,19 @@ async function createUser({
...
   // insert into DB after creation
-  const user = await User.create(userObject);
+  return await User.create(userObject);
+}
 
-  return user;
+async function createManyUsers(numberOfUsers: number): Promise<User[]> {
+  const userArray = Array.from({ length: numberOfUsers }, makeUserObject);
+  return await Db.sequelize.transaction(async transaction => {
+    return await User.bulkCreate(userArray, { transaction });
```
timmyichen, line 64: unnecessary return here - can just be `await User.bulkCreate ...`

server/test/utils.ts
timmyichen: `createUsers` being similar to `createUser` is completely fine if it makes sense semantically - as in, if it's just doing multiple of `createUser`, then i think it makes sense.

the issue that i had with makeUser/createUser is that, to those not familiar with Laravel  conventions (TIL!), the difference between the two is ambiguous.

with createUser/createUsers, though similar in the name itself, the meaning is different in that one is the pluralized form of the other

timmyichen (commented):
mostly looks good to me, just some final comments!

server/test/utils.ts
devonzara: I may be mistaken, but I do believe both `return`s are necessary. The opposite could be said for both `await`s though.

timmyichen: ah yup, you're right - misunderstood how these were getting returned.  thanks for clearing that up!

server/test/utils.ts
jmoldyvan: alrighty, no problemo

tsconfig.json
```diff
@@ -41,7 +41,7 @@
     "server/**/*",
     "client/**/*",
     "shared/**/*",
-    "pages/**/*"
+    "pages/**/*",
```
devonzara, line 44: Oops, overlooked this earlier... JSON should not have dangling commas.

devonzara (changes requested):

server/test/utils.ts
```diff
@@ -77,4 +75,6 @@ export {
   randomEmptyChance,
   createUser,
   getExpectedUserObject,
+  makeUserObject,
+  createUsers,
```
devonzara, line 79: nit: I typically sort exports 1 of 2 ways. Alphabetically, or in the order in which they appear above. I tend to prefer alphabetically.

tsconfig.json
timmyichen: fun fact, tsconfig jsons are not "real" json and can have comments, dangling commas, etc: https://github.com/microsoft/TypeScript/issues/20384#issuecomment-348552936

server/test/utils.ts
timmyichen: i feel like we can also just export the functions directly, e.g. `export function getExpectedUserObject(user: User) {`

devonzara: Then they'll have to be below all non-exported functions or we'll need to remove one of the ESLint rules I added.

tsconfig.json
devonzara: Ahh, interesting... wasn't aware of that. I do wish they used a `jsonc` file extension or something to indicate this difference. In that case, I guess it doesn't really matter. 

For the mere purpose of consistency throughout the file, I'd still remove it. Though no longer a required change. 👍🏼

devonzara (approved):

timmyichen (approved):

Caleb-Cohen (approved):

PR #81, opened 2023-04-02
#55 add cypress to automatic checks+optimizations

### LINKED ISSUE #55: Add Cypress to automatic checks + optimizations
**Related Issues**
Depends on: #54 

**Description**
Now that we're building the app as part of the GHA, we can run tests on the built test files rather than running it on TypeScript files.

Additionally, we can dockerize the app, run it in GHA, and then run cypress tests against the dockerized app.  This can probably replace the existing postgres docker stack that's being spun up for server tests.

Unfortunately, there's no good way to actually test things GHA other than just making a bunch of commits and seeing if it passes. 🙃 

**Acceptance Criteria**
* Cypress tests run as part of GHA
* Tests are run against a dockerized built version of the app

windyScripts:
Closes #55

# Description

The goal was to build and deploy the app using docker, following which linting and then server and client tests would be run on said built app.
* .gitignore was updated to include cypress/screenshots and cypress/videos
* A Dockerfile containing instructions to build the app and have it listen on port 3000 has been created
* A second docker compose file has been created to build the container from the Dockerfile, with .env variables specified.
* Github Actions workflow timeout minutes has been raised to 8 as the workflow takes more than five minutes.
* Github Actions workflow has been altered to accommodate the use of docker containers.

## Testing

The app should build and linting, server and client testing should happen on push or pull requests related to the main branch.

## Type of change

- Change to documentation/GitHub flows/CICD

# Checklist:


.github/workflows/lint.yml
```diff
@@ -40,21 +40,20 @@ jobs:
...
-      - name: Start Container
-        run: docker-compose -f "docker-compose.yml" up -d
 
-      - name: Run Tests
+      - name: Run Jest tests
         run: npm run test
 
+      - name: Cypress run
+        uses: cypress-io/github-action@v5.1.0 # use the explicit version number
+
```
devonzara, line 56: Looking good, but you need to build and start the server... something like this should get you going:
```suggestion
      - name: Cypress run
        uses: cypress-io/github-action@v5.1.0 # use the explicit version number
          build: npm run build
          start: npm start
          wait-on: 'http://localhost:3000'
```

There's lots of useful information on [cypress-io/github-action](https://github.com/cypress-io/github-action). 👍🏼

README.md
```diff
@@ -97,12 +97,15 @@ npm run test:watch -- auth.test.ts
 
 We use [cypress](https://docs.cypress.io/guides/overview/why-cypress) for UI testing. Cypress requires the app to be running for testing.
 
-To get started,
+To run tests manually,
 * Execute `npm run cypress:open`
 * Select E2E testing
 * Choose a browser
 * run the requisite test spec from the list.
```
devonzara, line 104: nit: Punctuation, capitalization, simplified grammar
```suggestion
To run tests manually:
* Execute `npm run cypress:open`
* Select E2E testing
* Choose a browser
* Select the test spec you wish to run
```

.github/workflows/lint.yml
windyScripts: I've both npm run build and npm start are part of the dockerfile, and run when it's called by the docker compose file. do I have to have to repeat the statements again?

devonzara: I'm uncertain as to why a Dockerfile is being used. Was there a particular need for it?

If we keep it, we'll need to edit the `npm` scripts to only start the `pg` service or create a separate docker-compose for the workflow.

If it's not needed and doesn't provide any additional benefit, I'd opt to do away with it and stick to the changes I suggested above.

timmyichen: ideally we'd use the built files for other things as well, such as running server tests.  i imagine we _could_ use the cypress build step for this but i think it'd be better to have the different test suites to work independently of each other, even if we can guarantee that we'll always run jest acceptance tests after cypress e2e tests.

devonzara: Fair... In that case, I guess we do need a second `docker-compose` file... one with the app build step.

We'll also need to add an environment variable for `DATABASE_URL` with the host configured as `pg`.

docker-compose.ci.yml
```diff
@@ -0,0 +1,14 @@
+version: "3.9"
+
+services:
+  app:
+    build: .
+    environment:
+      DB_HOST: localhost
```
devonzara, line 7: Since the app is also running in a container, I believe the `DB_HOST` should be set to `pg` (the name of the PostgreSQL service defined in `docker-compose.yml`.

windyScripts: While testing, I've found that `DB_HOST `value of pg leads to failure, while a value of dev-directory_pg_1 gets accepted. I'm unable to explain this. Would you know why this might be the case?

windyScripts:
@timmyichen  @devonzara the workflow is running without issues. Would you have any recommendations before making this an active PR?

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM node:18-alpine  
+
+WORKDIR /app
+
+COPY package*.json ./  
+
+RUN npm ci
+
+COPY . .
```
devonzara, line 9: Why are we doing a copy twice? Will it work if you do a single copy? eg.
```suggestion
COPY . .

RUN npm ci
```

.github/workflows/lint.yml
```diff
@@ -37,24 +37,24 @@ jobs:
         with:
           node-version: 18
 
-      - name: Install Node.js dependencies
-        run: npm ci
+      - name: Install Cypress
+        run: npm i cypress
```
timmyichen, line 41: is there a reason why this needs its own step? is it for the `run cypress tests` action down below?

devonzara (changes requested):
Few minor suggestions, but otherwise looks good.

.github/workflows/lint.yml
```diff
@@ -37,24 +37,24 @@ jobs:
...
+        run: docker-compose -f "docker-compose.ci.yml" up -d --build 
 
       - name: Run linters
         uses: wearerequired/lint-action@v2
         with:
           eslint: true
           eslint_extensions: "js,jsx,ts,tsx"
+      
+      - name: Run Jest tests
+        run: docker exec dev-directory_app_1 npm run test
```
timmyichen, line 53: does this run tests against the built files, or is it just running in the context of the app?

timmyichen (commented):
some minor comments, but this looks great! thanks for taking the time to hack this one out :)

.github/workflows/lint.yml
windyScripts: It is. Would it be better to integrate these two into one step?

.github/workflows/lint.yml
windyScripts: Have corrected this.

.github/workflows/lint.yml
windyScripts: Have corrected this

.github/workflows/lint.yml
```diff
@@ -37,24 +37,24 @@ jobs:
...
 
-      - name: Install Node.js dependencies
-        run: npm ci
+      - name: Install Cypress
+        run: npm i cypress
 
-      - name: Build App
-        run:  npm run build
+      - name: Build and serve App Container
+        run: docker-compose -f "docker-compose.ci.yml" up -d --build
```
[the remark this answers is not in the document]
windyScripts: I've made this correction

windyScripts:
@devonzara let me know if everything's good now.

docker-compose.ci.yml
```diff
@@ -0,0 +1,25 @@
...
+    environment:
+      DB_PASSWORD: ${DB_PASSWORD:-password}
+      DB_HOST: ${DB_HOST:-pg}
+      DB_PORT: ${DB_PORT:-5432}
+      DB_USER: ${DB_USER:-postgres}
+      DB_NAME: ${DB_NAME:-dev-directory}
+    ports:
+      - 3000:3000
+    depends_on:
+        - pg
```
devonzara, line 17: Indentation is off here.

devonzara (changes requested):
Small typo in the YAML. Also, are the env variables at the top of `lint.yml` still necessary?

docker-compose.ci.yml
windyScripts: Corrected this and removed the redundant ENV variables.

devonzara (approved):
Looks good to me!

I am a bit confused as to why Cypress and ESLint are run outside of the container while Jest is being run inside it. Also, I didn't realize you could (or should?) run the tests against the built project.

That said, if it works as intended (which, it appears to); I won't dispute it as that's outside of my area of expertise. 👍🏼

timmyichen:
> I am a bit confused as to why Cypress and ESLint are run outside of the container while Jest is being run inside it. Also, I didn't realize you could (or should?) run the tests against the built project.

Running tests on compiled code runs faster than raw TS, so we get a (potentially non trivial) speed boost by using the compiled code.

Running tasks in the container generally runs slower than outside the container.  Jest should probably be run outside the container as well, but I wouldn't call it a blocker either.

windyScripts:
I hadn't considered the implications of the environment on speeds for running jest tests, my thinking was along the lines that server tests should be run in the context of the container/built app, since I have limited experience with CI systems. I see why it's a better idea to move the server tests outside as well.

timmyichen (approved):
🎉 thanks for taking this beast on! :)

Issue #79, opened 2023-04-02
Add Redirect For Users That Have Been Through Onboarding


jmoldyvan:
**Related Issues**
- Project: #23
- Depends on: 
  - #78

**Description**
- Include the logic to redirect users to the dev-directory list page after login, based on the flags created in #78.

**Acceptance Criteria**
Users who have the specified flag from #78 will be redirected without seeing the onboarding.

jmoldyvan:
I can grab this once #78 is done

timmyichen:
This might actually [be done](https://github.com/timmyichen/dev-directory/commit/b37b5be350ebb578bc1a11a41176aab88bf0c08f#diff-ac7d80d71d40a2b128439c34595c6c71b51f562fd55ec73eaf7f358f6f00eeeeR18-R21), @jmoldyvan - let me know if you find otherwise!

jmoldyvan:
YUP! Looks like the functionality for skipping has been introduced! This issue is completed @timmyichen

Issue #78, opened 2023-04-02
Create Flags That Check If User Has Been Through Onboarding


jmoldyvan:
**Related Issues**
- Project: #23
- Depends on: 
  - #76

**Description**

Add an additional table `flags` that has the following schema:
* `user_id` foreign key to `users (id)`
* `flag_name` that represents the flag we are setting.

Once the user has skipped or completed onboarding, insert a new row in the `flags` tables, for example: `(user1, 'hasBeenOnboarded')`

Add additional routes for fetching a user's flags:
```js
const flags = await Flags.findAll({ where: { user_id: user.id } });
return flags.map(flag => flag.flag_name);
```

**Acceptance Criteria**
Make sure the tables are created.
Have the tables new rows created.
The new routes are functional.

DevinCLane:
I'll take a crack at this

Issue #59, opened 2023-03-24
Display user profiles returned from `/api/users` route


DevinCLane:
**Related Issues**
- Project: #17 
- Depends on: #58  

**Description**
- Display user profiles returned from `/api/users` route on home page.
- Show Discord names for each user.

**Acceptance Criteria**
- User profiles are shown on the home page.
- Acceptance tests exist for this feature.

timmyichen:
On further thought, we should probably create a separate page for this; being able to use the home page to introduce the app seems like something we'd want to do.  maybe `/directory`?

DevinCLane:
Hm yeah. I was imagining the first point of contact being the Login page that has a short description of what the app is/does, and then once you log in, you have the directory. Open to what you're describing though too.

Caleb-Cohen:
Client-side testing goes where?

devonzara:
Are we set on `/directory`? Thoughts on `/browse`, `/developers`, or even just `/search`?

timmyichen:
I'm fine with any of them; it can easily be changed from one to the other.

PR #82, opened 2023-04-04
Add Onboarding page and redirect from login via auth-redirect

### LINKED ISSUE #76: Create OnBoarding Page and Skip Functionality
**Related Issues**
- Project: #23
- Depends on: N/A

**Description**
- Use MUI to create a new page where users are redirected to, after they log in. 
- Create a dismiss button that will redirect users to the dev-directory list of profiles page.

**Acceptance Criteria**
The new page is created and the dismiss button is functional.
Redirection from the discord login is functional.
Redirection from onboarding to the dev-directory list of profile pages is functional.

MainlyColors:
Closes #76 

# Description

- add redirect from `/auth-redirect` to `/onboarding` instead of the previous `/`
- Add Onboarding page with skip button
- skip button takes you back to `/` (temporary till `/directory` exists )

## Testing

To Manually test
1. ` npm run dev:server` w/ docker already running
2. Login on `/` route
3. Authorize Discord
4. Will be redirected from oAuth page to `/auth-redirect`
5. On `/auth-redirect` route after verifying the user is authorized via a `router.query.code` existing, page will redirect to `/onboarding` instead of the previous `/`
6. verify you see the onboarding page
7. click skip
8. will be redirected to home page with this url `/` (temporary till `/directory` exists )
9. testing done

## Type of change

## Screenshots
`/onboarding/`
image

# Checklist:


pages/index.tsx
```diff
@@ -10,32 +11,50 @@ import { AUTH_COOKIE_NAME } from 'shared/constants';
...
+      setSeenOnBoarding(true);
+      // reset query parameters once user gets back to home page
+      router.replace({
+        pathname: '/',
+        query: {},
+      });
+    } else if (!seenOnBoarding && isAuthed) {
+      router.push('/onBoarding');
+    }
+  }, []);
```
MainlyColors, line 36: Highly influenced by ChatGPT 👀

Let me know if there is a better way, Next is confusing

jmoldyvan: i'll check it out today! thanks!

pages/index.tsx
```diff
@@ -53,9 +72,9 @@ Index.getInitialProps = async ({ req }) => {
     const axios = createAxiosInstance(req);
     await axios.get('/api/users/' + payload.user_id);
 
-    return { isAuthed: true };
+    return { isAuthed: true, seenOnBoarding: false };
   } catch (error) {
-    return { isAuthed: false };
+    return { isAuthed: false, seenOnBoarding: false };
```
timmyichen, line 77: is this necessary, given that it will always be false?

pages/index.tsx
```diff
@@ -10,32 +11,50 @@ import { AUTH_COOKIE_NAME } from 'shared/constants';
...
     const axios = createAxiosInstance();
     await axios.post('/api/auth/logout');
     setIsAuthed(false);
+    setSeenOnBoarding(false);
   };
 
   return (
-    <Container maxWidth="lg" className="flex justify-between pt-4">
-      <Typography variant="h1" className="text-3xl">
+    <Container maxWidth='lg' className='flex justify-between pt-4'>
```
timmyichen, line 46: why the quote change? JSX should use double quotes to imitate HTML conventions

pages/onBoarding/index.tsx
devonzara, line 1: Hyphenated lowercase URLs are typically preferred; therefore, `on-boarding` would be appropriate.

However, since "onboarding" is a single word, it should instead be `onboarding`.

pages/index.tsx
```diff
@@ -10,32 +11,50 @@ import { AUTH_COOKIE_NAME } from 'shared/constants';
...
 }
 
 const Index: NextPage<Props> = props => {
   const [isAuthed, setIsAuthed] = React.useState(props.isAuthed);
+  const [seenOnBoarding, setSeenOnBoarding] = React.useState(props.seenOnBoarding);
+  const router = useRouter();
+
+  // to avoid this error:
+  // During Pre-rendering (SSR or SSG) you tried to access a router method push, replace, back, which is not supported.
+  React.useEffect(() => {
```
timmyichen, line 24: What's being done to avoid the error?  `React.useEffect` only runs clientside so this should behave as expected?  What was being attempted before?

pages/onBoarding/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  };
+
+  return (
+    <Container maxWidth='lg' className='flex justify-between pt-4'>
+      <Typography variant='h1' className='text-3xl'>
+        OnBoarding
+      </Typography>
+      <Box className='flex items-center gap-4'>
+        <Button variant='contained' color='secondary' onClick={onDismiss}>
+          Dismiss
```
timmyichen, line 18: The term "dismiss" makes me think of modals, or things that are currently in your way - given this is a full page, I think "skip" is more appropriate

timmyichen (changes requested):
From the created ticket:
> Use MUI to create a new page where users are redirected to, after they log in.
> Create a dismiss button that will redirect users to the dev-directory list of profiles page.

I took this to mean:
* When the user logs in, instead of redirecting to the index page, we send them to onboarding page
* When the user skips onboarding, they get sent to the home page (i left a comment on #59 around changing where the actual directory lives)
* If they refresh while on the home page, they are not prompted to go through onboarding again

This PR does the following:
* Anyone who lands on the home page, regardless of whether they are authed or not, will be sent to the onboarding page
* When the user skips onboarding, they get sent back to the home page
* If they refresh while on the home page, they will be redirected to the onboarding page again

I think we may want to change this to be more like the former - when users log in via Discord, their code is validated on the `auth-redirect` page.  This is probably where we want to be making changes related to redirecting to onboarding.

General nit: "onboarding" is one word, so any instances of "onBoarding" or "xOnBoarding" should probably just be changed to "onboarding" and "xOnboarding" respectively

pages/index.tsx
devonzara: Be careful with scope creep. If I'm understanding #76 correctly, this ticket should redirect to onboarding after each login, unconditionally.

Keeping track of if the user has seen the onboarding page is not a part of this ticket. We need #78 to create the data structure to track that and #79 will come back to implement it.

_As a side note, this logic should probably have been in `client/pages/auth-redirect` since it determines where we go after a login and doesn't need executed everytime we hit the homepage._

pages/index.tsx
```diff
@@ -10,32 +11,50 @@ import { AUTH_COOKIE_NAME } from 'shared/constants';
...
+    } else if (!seenOnBoarding && isAuthed) {
+      router.push('/onBoarding');
+    }
+  }, []);
 
   const onLogout = async () => {
     const axios = createAxiosInstance();
     await axios.post('/api/auth/logout');
     setIsAuthed(false);
+    setSeenOnBoarding(false);
```
devonzara, line 42: See previous comment: Outside of scope.

pages/index.tsx
devonzara, line 77: See previous comment: Outside of scope.

pages/index.tsx
devonzara: Oh, didn't even consider adding some JSX rules to ESLint... 🤦🏼‍♂️😅

devonzara (commented):
Looks good, just does a bit more than asked. Also, JSX should default to double quotes. 👍🏼

pages/index.tsx
timmyichen: was this your linting change 🥲

devonzara:
Oof, I wish the review page would update to show other reviewer comments as they post them like it does when there's a new commit... Tim hit on essentially all the same points; I could have been watching TV instead. 😅

pages/index.tsx
devonzara: No, it wasn't! I actually panicked when I saw his changes, went and double-checked! 😂

jmoldyvan:
I got caught up with some stuff, but I will look on my machine as soon as I can

pages/index.tsx
MainlyColors: My bad, I didn't realize this was scope creep at all, I thought I was doing the bare minimum to get the feature working. 

I'll make the changes, I didn't understand how the auth-redirect route was getting hit but ill look into it more.

pages/index.tsx
MainlyColors: I tried `router.push('/onBoarding');` without `useEffect` hook first and got the error. So I added the comment about the error because i thought it was a workaround but if its normal ill remove it

pages/index.tsx
MainlyColors: My thought process was in the future we would have more scenarios where we want to change `seenOnboarding` state but for now I was just putting false on both because i didn't know what those scenarios would be yet

I removed both, since I know how auth-redirect page was coming into the mix, and the code is better served being there.

pages/index.tsx
MainlyColors: My bad, prettier was set to single quotes for jsx, didn't know it was a convention to use double quotes with JSX

pages/onBoarding/index.tsx
MainlyColors: Agreed, fixed

pages/index.tsx
MainlyColors: removed

pages/index.tsx
MainlyColors: removed

pages/index.tsx
MainlyColors: no longer needed, removed

DevinCLane:
Also if we're switching to `onboarding` from `onBoarding`, let's update the directory name to `/pages/onboarding` instead of the current `/pages/onBoarding`

MainlyColors:
> Also if we're switching to `onboarding` from `onBoarding`, let's update the directory name to `/pages/onboarding` instead of the current `/pages/onBoarding`

Thought I did that before because it changed on my end but I guess there is a specific git way to do a rename lol

Fixed now

pages/index.tsx
timmyichen: To clarify, the error is expected in the sense that routing logic shouldn't be happening at the top level of the component. The main body of a functional component is evaluated on every render.

pages/onboarding/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  return (
+    <Container maxWidth="lg" className="flex justify-between pt-4">
+      <Typography variant="h1" className="text-3xl">
+        OnBoarding
+      </Typography>
+      <Box className="flex items-center gap-4">
+        <Button variant="contained" color="secondary" onClick={onSkip}>
+          Skip
+        </Button>
```
devonzara, line 19: I don't believe the click handler is necessary. Instead, we can just wrap the `Button` with a NextJS `Link` component ([docs](https://nextjs.org/docs/api-reference/next/link)).

```suggestion
        <Link href="/" passHref>
          <Button variant="contained" color="secondary">Skip</Button>
        </Link>
```

devonzara (changes requested):
👍🏼 Few additional touch-ups.

pages/onboarding/index.tsx
MainlyColors: Yea a button should do something on the page, a link takes you somewhere. So semantically it makes sense

Fixed

devonzara:
nit: might be worth editing the PR's title, but otherwise looks good. 👍🏼

pages/onboarding/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { Box, Button, Container, Typography } from '@mui/material';
+import { NextPage } from 'next';
+import Link from 'next/link';
+
+const Index: NextPage = () => {
```
timmyichen, line 5: nit: rename `Index` to `Onboarding`

pages/onboarding/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <Container maxWidth="lg" className="flex justify-between pt-4">
+      <Typography variant="h1" className="text-3xl">
+        Onboarding
+      </Typography>
+      <Box className="flex items-center gap-4">
+        <Link href="/" passHref>
```
timmyichen, line 12: what's the purpose of the `passHref` prop here? From the docs it says:

> Forces Link to send the href property to its child. Defaults to false

But a button shouldn't need an href?

timmyichen (commented):
Mostly LGTM, just some small comments. Thanks for making the requested changes!

pages/onboarding/index.tsx
devonzara: It shouldn't really be a button to begin with, it's linking to another page... it's not (currently) doing any background tasks on the current page. MUI's button accepts a `href` ([docs](https://mui.com/material-ui/api/button/#props)).
> href | string | The URL to link to when the button is clicked. If defined, an a element will be used as the root node.

This is basically making it a link while keeping it styled like a button.

timmyichen: ah, gotcha. thanks for explaining!

MainlyColors: So I tested with and without, In both cases the page transfer still activates but the HTML is rendered differently

This is without `passHref`

```html
<div class="flex items-center gap-4 MuiBox-root css-0">
<button class="(...many classes)" tabindex="0" type="button">
Skip
<span class="MuiTouchRipple-root css-8je8zh-MuiTouchRipple-root"></span>
</button>
</div>
```

This is with `passHref`

```html
<div class="flex items-center gap-4 MuiBox-root css-0">
<a class="(...many classes)" tabindex="0" href="/">
Skip
<span class="MuiTouchRipple-root css-8je8zh-MuiTouchRipple-root"></span>
</a>
</div>
```

timmyichen (approved):

pages/onboarding/index.tsx
devonzara: Correct, `passHref` on the NextJS's `Link` component is passing `href` into its children; which, in this case, is the MUI `Button`.

The MUI button will then change the "root node" to an `a` instead of a `button` element when `href` is passed into it. Thus resulting in the changes you observed.

PR #85, opened 2023-04-07
Visualize users

### LINKED ISSUE #59: Display user profiles returned from `/api/users` route

Caleb-Cohen:
Closes #59 

# Description

Added visualizing users onto the homepage. Only the discord name is visualized per the issue request.

## Testing

Describe what a reviewer should do to test your code, including instructions for manually testing as well as relevant automated test suites to run.

execute `npm run cypress:open` and navigate the GUI to start the test.

## Type of change

## Screenshots

Screenshot 2023-04-06 at 10 28 58 PM

# Checklist:

### DIFF
--- a/cypress/e2e/base.spec.ts
+++ b/cypress/e2e/base.spec.ts
@@ -10,4 +10,14 @@ describe('Open browser and check if login button exists.', () => {
   });
 });
 
+describe('Visit directory page and verify users exists', () => {
+  beforeEach(() => {
+    cy.visit('http://localhost:3000/directory');
+  });
+
+  it('Displays a list of users', () => {
+    cy.get('ul').should('be.visible');
+  });
+});
+
 export {};
--- a/pages/directory/index.tsx
+++ b/pages/directory/index.tsx
@@ -6,8 +6,8 @@ import createAxiosInstance from 'client/lib/axios';
 
 //TODO: Type users once sequelize-typescript is added
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const Directory: NextPage<{ users: any[] }> = props => {
-  const { users } = props;
+const Directory: NextPage<{ users: any[]; error: string }> = props => {
+  const { users, error } = props;
 
   return (
     <Container maxWidth="lg" className="pt-4">
@@ -16,6 +16,7 @@ const Directory: NextPage<{ users: any[] }> = props => {
           Users:
         </Typography>
         <Box>
+          {error && <Typography>{error}</Typography>}
           <List>
             {users.map(user => (
               <ListItem key={user.id}>{user.discord_name}</ListItem>
@@ -33,9 +34,9 @@ Directory.getInitialProps = async ({ req }) => {
     const response = await axios.get('/api/users');
     const users = response.data.users;
 
-    return { users };
+    return { users, error: null };
   } catch (error) {
-    return error;
+    return { users: [], error: error.message };
   }
 };
 
--- a/pages/directory/index.tsx
+++ b/pages/directory/index.tsx
@@ -28,14 +28,14 @@ const Directory: NextPage<{ users: any[] }> = props => {
 };
 
 Directory.getInitialProps = async ({ req }) => {
-  const axios = createAxiosInstance(req);
-  const response = await axios.get('/api/users');
-  const users = response.data.users;
-
   try {
+    const axios = createAxiosInstance(req);
+    const response = await axios.get('/api/users');
+    const users = response.data.users;
+
     return { users };
   } catch (error) {
-    return { users };
+    return error;
   }
 };
 
--- a/pages/directory/index.tsx
+++ b/pages/directory/index.tsx
@@ -1,0 +1,42 @@
+import { Box, Container, List, ListItem, Typography } from '@mui/material';
+import { NextPage } from 'next';
+import React from 'react';
+
+import createAxiosInstance from 'client/lib/axios';
+
+//TODO: Type users once sequelize-typescript is added
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const Directory: NextPage<{ users: any[] }> = props => {
+  const { users } = props;
+
+  return (
+    <Container maxWidth="lg" className="pt-4">
+      <Box className="pt-4">
+        <Typography variant="h2" className="text-2xl">
+          Users:
+        </Typography>
+        <Box>
+          <List>
+            {users.map(user => (
+              <ListItem key={user.id}>{user.discord_name}</ListItem>
+            ))}
+          </List>
+        </Box>
+      </Box>
+    </Container>
+  );
+};
+
+Directory.getInitialProps = async ({ req }) => {
+  const axios = createAxiosInstance(req);
+  const response = await axios.get('/api/users');
+  const users = response.data.users;
+
+  try {
+    return { users };
+  } catch (error) {
+    return { users };
+  }
+};
+
+export default Directory;


cypress/e2e/base.spec.ts
```diff
@@ -10,4 +10,23 @@ describe('Open browser and check if login button exists.', () => {
   });
 });
 
+describe('Visit landing page and verify users exists', () => {
+  beforeEach(() => {
+    cy.intercept('/api/users', { fixture: 'users.json' }).as('getUsers');
```
Caleb-Cohen, line 15: I'm not confident about my attempt to intercept `/api/users` with data that doesn't exactly match what is being sent. The reason I kept it simplified is I expect what is sent to the client to change a fair bit of times throughout the process, and I was cautious about creating a response much more than what was necessary for this specific issue.

pages/index.tsx
```diff
@@ -1,19 +1,36 @@
...
 import Cookies from 'universal-cookie';
 
 import createAxiosInstance from 'client/lib/axios';
 import { getDiscordOauthUrl } from 'client/lib/oauth';
 import { AUTH_COOKIE_NAME } from 'shared/constants';
 
+interface User {
+  id: number;
+  discord_name: string;
+}
```
Caleb-Cohen, line 14: We've talked about interface's not being necessary in some cases. It's unclear to me if it's needed here.

Caleb-Cohen (commented):
Added some comments

pages/index.tsx
```diff
@@ -1,19 +1,36 @@
...
   const [isAuthed, setIsAuthed] = React.useState(props.isAuthed);
+  const [users, setUsers] = useState<User[]>([]);
+
+  useEffect(() => {
+    const fetchUsers = async () => {
+      const axios = createAxiosInstance();
+      const response = await axios.get('/api/users');
+      const users = response.data.users;
+      setUsers(users);
+      console.log(users);
```
Caleb-Cohen, line 30: Who would leave such a thing here

pages/index.tsx
```diff
@@ -1,19 +1,36 @@
...
+    };
+    fetchUsers();
+  }, []);
```
timmyichen, line 33: Making the call in `useEffect` will ensure that the API request only happens client-side, meaning we aren't taking advantage of Next's SSR capabilities. This should probably be moved to `getInitialProps`

pages/index.tsx
timmyichen: This should probably also be a shared type between the server and the client so that we can reference the "client shape User" both:
* when we make the call for a user on the client, and
* when we return it on the server

pages/index.tsx
```diff
@@ -1,19 +1,36 @@
 import { Box, Button, Container, Typography } from '@mui/material';
 import jwt from 'jsonwebtoken';
 import { NextPage } from 'next';
-import React from 'react';
+import React, { useState, useEffect } from 'react';
```
timmyichen, line 4: Personal preference, but I think it's easier to understand React hook usage when we _don't_ import hooks, and instead use them as `React.useState()` or `React.useEffect()`.  This allows us to distinguish between a custom hook (which will also be named as `useX`) and an official React hook. Non-blocking comment though!

pages/index.tsx
```diff
@@ -22,23 +39,37 @@ const Index: NextPage<Props> = props => {
...
+              Log out
+            </Button>
+          ) : (
+            <Button variant="contained" color="primary" href={getDiscordOauthUrl()}>
+              Log in
+            </Button>
+          )}
+        </Box>
+      </Box>
+      <Box sx={{ pt: 4 }}>
```
timmyichen, line 62: Is there a reason this uses the `sx` prop vs the tailwind class name (as is used in the `Container` component above)?

cypress/fixtures/users.json
```diff
@@ -0,0 +1,126 @@
+{
+    "page": 1,
+    "totalPages": 1,
+    "users": [
```
timmyichen, line 4: As much as I enjoy the test data here 🙃 it may be a hassle to update this when we start passing more fields back to the client (i.e. all the other fields on `User`) - can we instead generate this using some of the test utils that we've created for the server?

cypress/e2e/base.spec.ts
timmyichen, line 15: The `getUsers` alias isn't being used anywhere - we can probably remove

timmyichen: Also, general note for anyone: we may not be able to intercept API calls that are called on the server (as a result of SSR) - definitely worth looking into whether it's possible.

If it's not easily implemented, I think it's fine if we just don't test that particular functionality as a part of this PR

timmyichen (commented):
Thanks for taking this on! Left a few comments, largely around how we can use SSR to improve performance. As a side note, per [this comment](https://github.com/timmyichen/dev-directory/issues/59#issuecomment-1496868673) on the issue, can we move the added logic to a new page?

pages/index.tsx
```diff
@@ -22,23 +39,37 @@ const Index: NextPage<Props> = props => {
...
-          <Button variant="contained" color="primary" href={getDiscordOauthUrl()}>
-            Log in
-          </Button>
-        )}
+        <Box>
+          <ul>
+            {users.map(user => (
+              <li key={user.id}>{user.discord_name}</li>
+            ))}
+          </ul>
```
devonzara, line 71: Should be making use of MUI components for the `ul`/`li`.

cypress/e2e/base.spec.ts
Caleb-Cohen: What's SSR

pages/index.tsx
```diff
@@ -1,36 +1,17 @@
...
-  id: number;
-  discord_name: string;
-}
-
-interface Props {
-  isAuthed: boolean;
-}
-
-const Index: NextPage<Props> = props => {
+const Index: NextPage<{ isAuthed: boolean; users: User[] }> = props => {
```
Caleb-Cohen, line 12: @timmyichen Can I get some clarification on how you wanted the User to be referenced in this?

devonzara (approved):

devonzara:
Looks good to me. 👍🏼

pages/index.tsx
timmyichen: Do we have any types representing the a user, but only the fields the client will see?  That's what we should be using here (we should avoid doing any kind of server import on the client)

devonzara: `UserProfile`, though it's in `server/types/User` [[Link](https://github.com/timmyichen/dev-directory/blob/75cc63466ce7b3cf4fcbb607c3b69d1b92054c74/server/types/User.ts#L3)]

timmyichen: guh.  more reason to use sequelize-typescript imo.  i'd be fine with just typing this as `any` and adding a TODO to type it once sequelize-typescript is added

Caleb-Cohen: > Do we have any types representing the a user, but only the fields the client will see? That's what we should be using here (we should avoid doing any kind of server import on the client)

What's wrong with importing types from the server?

timmyichen: Because the client should be assumed to _not_ have direct access to server code and resources.

The risk is that server code is now bundled with the client, exposing how the app works and allowing clients to get a sense of how the backend works and find vulnerabilities. Obviously this doesn't matter since it's open source, but it's general good practice to follow.

I suspect it's not actually breaking anything right now because `User` is only used as a type, which gets stripped out on compilation, but if we were to use `User` in any other non-type capacity it will probably break the app.

timmyichen (changes requested):
One last thing, can we move the users list to a new page `/directory`?

Caleb-Cohen:
> One last thing, can we move the users list to a new page `/directory`?

Do you want anything other than the users list on the directory page? Like `Dev-Directory` or log in stuff?

pages/index.tsx
Caleb-Cohen: How will sequelize-typescript give the client the type without having the same issue?

timmyichen: sequelize-typescript allows us to define an interface (which can be used by both client and server, and can live in the shared directory) and create a class that is coupled with the interface, allowing us to have a single source of truth for what fields exist for a user.

pages/directory/index.tsx
```diff
@@ -0,0 +1,42 @@
+import { Box, Container, List, ListItem, Typography } from '@mui/material';
+import { NextPage } from 'next';
+import React from 'react';
+
+import createAxiosInstance from 'client/lib/axios';
+
+//TODO: Type users once sequelize-typescript is added
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const Directory: NextPage<{ users: any[] }> = props => {
```
Caleb-Cohen, line 9: Lint did not like users: any. 

I added an ignore as this should be a temporary bandaid until sequilize-typscript

timmyichen:
Where did we land on putting this in a new page? We know for sure it'll be a different page, what that page is named is up in the air (I'm fine with anything), but we should move everything out of the index file since doing so will allow to actually iterate on the homepage. Otherwise we'll keep piling features on the homepage, making the separation in the future more hairy.

Caleb-Cohen:
> Where did we land on putting this in a new page? We know for sure it'll be a different page, what that page is named is up in the air (I'm fine with anything), but we should move everything out of the index file since doing so will allow to actually iterate on the homepage. Otherwise we'll keep piling features on the homepage, making the separation in the future more hairy.

Not sure I'm following the question. I moved all the related user information to the directory folder and modeled the structure similar to onboarding.

pages/index.tsx
devonzara, line 1: We should probably revert this file to its initial state as these changes aren't important or relevant to the issue.

devonzara (changes requested):
`pages/index.tsx` probably shouldn't have any changes.
Aside from that, looks good. 👍🏼

`/directory` is fine for now, we can change the URL later if we decide to...

pages/index.tsx
Caleb-Cohen: I do not understand what to do.

Caleb-Cohen: Ah, now I see. Fixed.

devonzara: 🤔 What happened to the commit history?

I mean, the changes to the file are gone, so... 👍🏼

timmyichen (approved):
caleb cohen and the case of the mysterious commit history

pages/directory/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+Directory.getInitialProps = async ({ req }) => {
+  const axios = createAxiosInstance(req);
+  const response = await axios.get('/api/users');
+  const users = response.data.users;
+
+  try {
+    return { users };
+  } catch (error) {
+    return { users };
+  }
```
devonzara, line 39: What's happening here? 🤔

If the fetch fails, we're not going to catch it since it's not in the `try` block and we're then returning the same thing in the `catch` case?

Was this the same in prior commits or did it change?

devonzara (changes requested):
One more thing...

pages/directory/index.tsx
timmyichen: oh good catch, i clearly did not look at this closely LOL

Caleb-Cohen: > What's happening here? 🤔
> 
> If the fetch fails, we're not going to catch it since it's not in the `try` block and we're then returning the same thing in the `catch` case?
> 
> Was this the same in prior commits or did it change?

the axios, response, users was always outside the try. Only the auth stuff was within the try catch

pages/directory/index.tsx
```diff
@@ -28,14 +28,14 @@ const Directory: NextPage<{ users: any[] }> = props => {
...
-
   try {
+    const axios = createAxiosInstance(req);
+    const response = await axios.get('/api/users');
+    const users = response.data.users;
+
     return { users };
   } catch (error) {
-    return { users };
+    return error;
```
devonzara: This is why loosely typing things can be problematic...

Our `Directory` component is expecting to receive an object with the structure `{ users: any[] }`. If the API returns an error, the `catch` case is returning the `error` object instead of a structure the component is expecting.

devonzara (changes requested):
The return type doesn't match what the `Directory` component expects...

pages/directory/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+Directory.getInitialProps = async ({ req }) => {
+  try {
+    const axios = createAxiosInstance(req);
+    const response = await axios.get('/api/users');
+    const users = response.data.users;
+
+    return { users };
+  } catch (error) {
+    return error;
```
timmyichen, line 38: props should always be objects, but we should also aim to make them as consistent as possible.  might consider the following shape?

```ts
interface Props {
  users?: any[];
  error?: string;
}
```

then you can render an error if `users` is falsy or `error` is truthy

PR #87, opened 2023-04-14
#77 Create Form in Onboarding Page

### LINKED ISSUE #77: Create The Form That Goes Into The New OnBoarding Page And Handle Submission
**Related Issues**
- Project: #23
- Depends on: 
  - #76

**Description**
- Create a form, in the new OnBoarding page, with each of the fields from the user profile. This will include the following: `bio`, `twitter_username`, `linkedin_url`, `github_username`, `website`
 - `bio` should have (string, max 1000 chars), other fields should have (string, max 200 chars)

- Create a submit button on the page that updates the user profile fields with what they have completed, using the `PATCH /api/users/:id` endpoint.
- Redirects users to the dev-directory list page after submission

**Acceptance Criteria**
User can complete and submit the form.
The user model is updated.
The user is successfully redirected.

Includes end-to-end tests.

jmoldyvan:
Closes #77 

# Description

This PR creates a form in the onboarding page and submits user information to be included in the database/profile of the user.

## Testing

completed form and checked the profile url to see if it has appeared and is correct

E2E cypress testing complete and successful 

## Type of change

# Checklist:

### DIFF
--- a/cypress/e2e/base.spec.ts
+++ b/cypress/e2e/base.spec.ts
@@ -20,31 +20,4 @@ describe('Visit directory page and verify users exists', () => {
   });
 });
 
-describe('Onboarding Page', () => {
-  afterEach(() => {
-    cy.truncateDatabase();
-  });
-
-  it('displays the correct heading', () => {
-    cy.visit('http://localhost:3000/onboarding');
-    cy.contains('h1', 'Onboarding');
-  });
-
-  it('fills out the form successfully', () => {
-    cy.visit('http://localhost:3000');
-    cy.login();
-    cy.visit('http://localhost:3000/onboarding');
-    cy.get('form').within(() => {
-      cy.get('textarea[name="bio"]').type('Lorem ipsum dolor sit amet.');
-      cy.get('input[name="twitter_username"]').type('test_twitter');
-      cy.get('input[name="linkedin_url"]').type('https://www.linkedin.com/in/test_linkedin');
-      cy.get('input[name="github_username"]').type('test_github');
-      cy.get('input[name="website"]').type('https://www.testwebsite.com');
-      cy.visit('http://localhost:3000/directory');
-    });
-
-    cy.url().should('equal', 'http://localhost:3000/directory');
-  });
-});
-
 export {};
--- a/cypress/e2e/base.spec.ts
+++ b/cypress/e2e/base.spec.ts
@@ -20,4 +20,27 @@ describe('Visit directory page and verify users exists', () => {
   });
 });
 
+describe('Onboarding Page', () => {
+  beforeEach(() => {
+    cy.visit('http://localhost:3000/onboarding');
+  });
+
+  it('displays the correct heading', () => {
+    cy.contains('h1', 'Onboarding');
+  });
+
+  it('fills out the form successfully', () => {
+    cy.get('form').within(() => {
+      cy.get('textarea[name="bio"]').type('Lorem ipsum dolor sit amet.');
+      cy.get('input[name="twitter_username"]').type('test_twitter');
+      cy.get('input[name="linkedin_url"]').type('https://www.linkedin.com/in/test_linkedin');
+      cy.get('input[name="github_username"]').type('test_github');
+      cy.get('input[name="website"]').type('https://www.testwebsite.com');
+      cy.visit('http://localhost:3000/');
+    });
+
+    cy.url().should('equal', 'http://localhost:3000/');
+  });
+});
+
 export {};
--- a/cypress/e2e/base.spec.ts
+++ b/cypress/e2e/base.spec.ts
@@ -21,30 +21,19 @@ describe('Visit directory page and verify users exists', () => {
 });
 
 describe('Onboarding Page', () => {
-  
-  beforeEach(() => {
-    cy.intercept('http://localhost:3000/api/users/*', (req) => {
-      req.reply({
-        data: {
-          bio: "Lorem ipsum dolor sit amet",
-          discord_name: "johndoe#1111",
-          discord_user_id: "390925812364345344",
-          github_username: "johndoe",
-          id: 201,
-          linkedin_url: "https://www.johndoe.com",
-          twitter_username: "johndoe24",
-          website: "https://www.johndoe.com"
-        },
-      });
-    });
-    cy.visit('http://localhost:3000/onboarding');
+  afterEach(() => {
+    cy.truncateDatabase();
   });
 
   it('displays the correct heading', () => {
+    cy.visit('http://localhost:3000/onboarding');
     cy.contains('h1', 'Onboarding');
   });
 
   it('fills out the form successfully', () => {
+    cy.visit('http://localhost:3000');
+    cy.login();
+    cy.visit('http://localhost:3000/onboarding');
     cy.get('form').within(() => {
       cy.get('textarea[name="bio"]').type('Lorem ipsum dolor sit amet.');
       cy.get('input[name="twitter_username"]').type('test_twitter');
--- a/cypress/e2e/base.spec.ts
+++ b/cypress/e2e/base.spec.ts
@@ -21,7 +21,22 @@ describe('Visit directory page and verify users exists', () => {
 });
 
 describe('Onboarding Page', () => {
+  
   beforeEach(() => {
+    cy.intercept('http://localhost:3000/api/users/*', (req) => {
+      req.reply({
+        data: {
+          bio: "Lorem ipsum dolor sit amet",
+          discord_name: "johndoe#1111",
+          discord_user_id: "390925812364345344",
+          github_username: "johndoe",
+          id: 201,
+          linkedin_url: "https://www.johndoe.com",
+          twitter_username: "johndoe24",
+          website: "https://www.johndoe.com"
+        },
+      });
+    });
     cy.visit('http://localhost:3000/onboarding');
   });
 
--- a/cypress/e2e/base.spec.ts
+++ b/cypress/e2e/base.spec.ts
@@ -36,10 +36,10 @@ describe('Onboarding Page', () => {
       cy.get('input[name="linkedin_url"]').type('https://www.linkedin.com/in/test_linkedin');
       cy.get('input[name="github_username"]').type('test_github');
       cy.get('input[name="website"]').type('https://www.testwebsite.com');
-      cy.visit('http://localhost:3000/');
+      cy.visit('http://localhost:3000/directory');
     });
 
-    cy.url().should('equal', 'http://localhost:3000/');
+    cy.url().should('equal', 'http://localhost:3000/directory');
   });
 });
 
--- a/cypress/e2e/onboarding.spec.ts
+++ b/cypress/e2e/onboarding.spec.ts
@@ -1,17 +1,9 @@
 /// <reference types="cypress" />
 
 describe('Onboarding Page', () => {
-  afterEach(() => {
-    cy.truncateDatabase();
-  });
-
-  it('displays the correct heading', () => {
+  it('fills out the form successfully', () => {
     cy.login();
     cy.visit('http://localhost:3000/onboarding');
-    cy.contains('h1', 'Onboarding');
-  });
-
-  it('fills out the form successfully', () => {
     cy.get('form').within(() => {
       cy.get('textarea[name="bio"]').type('Lorem ipsum dolor sit amet.');
       cy.get('input[name="twitter_username"]').type('test_twitter');
@@ -22,6 +14,10 @@ describe('Onboarding Page', () => {
     });
 
     cy.url().should('equal', 'http://localhost:3000/directory');
+  });
+
+  after(() => {
+    cy.truncateDatabase();
   });
 });
 
--- a/cypress/e2e/onboarding.spec.ts
+++ b/cypress/e2e/onboarding.spec.ts
@@ -1,6 +1,10 @@
 /// <reference types="cypress" />
 
 describe('Onboarding Page', () => {
+  after(() => {
+    cy.truncateDatabase();
+  });
+
   it('fills out the form successfully', () => {
     cy.login();
     cy.visit('http://localhost:3000/onboarding');
@@ -14,10 +18,6 @@ describe('Onboarding Page', () => {
     });
 
     cy.url().should('equal', 'http://localhost:3000/directory');
-  });
-
-  after(() => {
-    cy.truncateDatabase();
   });
 });
 
--- a/cypress/e2e/onboarding.spec.ts
+++ b/cypress/e2e/onboarding.spec.ts
@@ -1,0 +1,28 @@
+/// <reference types="cypress" />
+
+describe('Onboarding Page', () => {
+  afterEach(() => {
+    cy.truncateDatabase();
+  });
+
+  it('displays the correct heading', () => {
+    cy.login();
+    cy.visit('http://localhost:3000/onboarding');
+    cy.contains('h1', 'Onboarding');
+  });
+
+  it('fills out the form successfully', () => {
+    cy.get('form').within(() => {
+      cy.get('textarea[name="bio"]').type('Lorem ipsum dolor sit amet.');
+      cy.get('input[name="twitter_username"]').type('test_twitter');
+      cy.get('input[name="linkedin_url"]').type('https://www.linkedin.com/in/test_linkedin');
+      cy.get('input[name="github_username"]').type('test_github');
+      cy.get('input[name="website"]').type('https://www.testwebsite.com');
+      cy.get('button[type="submit"]').click();
+    });
+
+    cy.url().should('equal', 'http://localhost:3000/directory');
+  });
+});
+
+export {};
--- a/pages/onboarding/index.tsx
+++ b/pages/onboarding/index.tsx
@@ -4,7 +4,7 @@ import jwt from 'jsonwebtoken';
 import { NextPage } from 'next';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import Cookies from 'universal-cookie';
 
 import createAxiosInstance from 'client/lib/axios';
@@ -25,11 +25,18 @@ interface Props {
 
 const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
   const [formData, setFormData] = useState<FormData>({
-    bio: user.bio,
-    twitter_username: user.twitter_username,
-    linkedin_url: user.linkedin_url,
-    github_username: user.github_username,
-    website: user.website,
+    bio: user?.bio,
+    twitter_username: user?.twitter_username,
+    linkedin_url: user?.linkedin_url,
+    github_username: user?.github_username,
+    website: user?.website,
+  });
+  const [originalFormData, setOriginalFormData] = useState<FormData>({
+    bio: user?.bio,
+    twitter_username: user?.twitter_username,
+    linkedin_url: user?.linkedin_url,
+    github_username: user?.github_username,
+    website: user?.website,
   });
   const [isLoading, setIsLoading] = useState<boolean>(false);
   const [isFormFilled, setIsFormFilled] = useState<boolean>(false);
@@ -39,7 +46,14 @@ const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
   const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     const { name, value } = event.target;
     setFormData(prevFormData => ({ ...prevFormData, [name]: value }));
-    setIsFormFilled(true);
+    console.log(formData);
+    console.log(originalFormData);
+    
+    if( user !== formData && originalFormData !== formData){
+      setIsFormFilled(true);
+    }
+    else if (originalFormData === formData){setIsFormFilled(false)}
+    
   };
 
   const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
@@ -47,7 +61,7 @@ const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
 
     try {
       setIsLoading(true);
-      await axios.patch(`/api/users/${user.id}`, formData);
+      await axios.patch(`/api/users/${user?.id}`, formData);
       setIsLoading(false);
       router.push('/directory');
     } catch (error) {
@@ -55,10 +69,6 @@ const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
       console.error(error);
     }
   };
-
-  if (!user) {
-    router.push('/');
-  }
 
   const isSubmitDisabled = !isFormFilled && !isLoading;
 
@@ -118,7 +128,6 @@ const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
         <Box mt={4} textAlign="right">
           <Button variant="contained" color="primary" type="submit" disabled={isSubmitDisabled}>
             {isLoading ? 'Submitting...' : 'Submit'}
-            Submit
           </Button>
         </Box>
       </form>
@@ -142,7 +151,14 @@ OnboardingPage.getInitialProps = async ({ req }) => {
     const userInfo = await axiosInstance.get('/api/users/' + payload.user_id);
     return { user: userInfo.data };
   } catch (error) {
-    return { user: null };
+    return { user:   {bio: "Lorem ipsum dolor sit amet",
+    discord_name: "johndoe#1111",
+    discord_user_id: "390925812364345344",
+    github_username: "johndoe",
+    id: 201,
+    linkedin_url: "https://www.johndoe.com",
+    twitter_username: "johndoe24",
+    website: "https://www.johndoe.com"} };
   }
 };
 
--- a/pages/onboarding/index.tsx
+++ b/pages/onboarding/index.tsx
@@ -1,4 +1,5 @@
-import { Box, Button, Container, TextField, Typography } from '@mui/material';
+import { Box, Button, Container, TextField, Typography, Snackbar } from '@mui/material';
+import MuiAlert, { AlertProps } from '@mui/material/Alert';
 import axios from 'axios';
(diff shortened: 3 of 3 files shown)


Caleb-Cohen:
WHERE TESTING

jmoldyvan:
> WHERE TESTING

HOL UP NEED TO MAKE IT BRB

jmoldyvan:
alright got the testing

pages/onboarding/index.tsx
windyScripts, line 1: nit: Change UseridData to UserIdData, fits convention and is (slightly) more readable.

windyScripts (approved):
Beyond the nit, LGTM

Caleb-Cohen:
Will add more comments later, but does the test confirm that the user profile is updated?

pages/onboarding/index.tsx
devonzara: I'd probably opt for `Props` or something instead, since it represents more than just the User that is held within.

pages/onboarding/index.tsx
```diff
@@ -1,8 +1,60 @@
...
+}
+
+const OnboardingPage: NextPage<UseridData> = ({ user }: UseridData) => {
+  const [formData, setFormData] = useState<FormData>({
+    bio: '',
+    twitter_username: '',
+    linkedin_url: '',
+    github_username: '',
+    website: '',
+  });
```
timmyichen, line 33: Initializing all these to be empty string will overwrite any existing data on the onboarding page - the default values here should be taken from the existing user data.

pages/onboarding/index.tsx
```diff
@@ -1,8 +1,60 @@
...
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    try {
+      const response = await axios.patch(`/api/users/${user.id}`, formData);
+
+      if (response.status !== 200) {
+        throw new Error('Failed to update user profile');
+      }
```
timmyichen, line 50: 2xx status codes are technically okay, 4xx and above are errors.  Additionally, the `axios.patch` call above will error if the request failed, so this error is unreachable.

pages/onboarding/index.tsx
```diff
@@ -1,8 +1,60 @@
...
+
+      router.push('/');
```
timmyichen, line 52: i think we want to redirect them to `/directory` instead? rather than back to the home page

pages/onboarding/index.tsx
```diff
@@ -15,8 +67,77 @@ const Onboarding: NextPage = () => {
           </Button>
```
timmyichen, line 67: i think it would also be best to put the skip button at the end of the form, rather than at the top - you'll see less user engagement if you present them with the option to skip right away

pages/onboarding/index.tsx
```diff
@@ -15,8 +67,77 @@ const Onboarding: NextPage = () => {
...
+        <TextField
+          label="Bio"
+          name="bio"
+          value={formData.bio}
+          onChange={handleChange}
+          multiline
+          rows={4}
+          fullWidth
+          inputProps={{ maxLength: 1000 }}
+          required
```
timmyichen, line 81: all these fields are required, but i'm not convinced they should be? e.g. why should a user be _required_ to enter a twitter username?

pages/onboarding/index.tsx
```diff
@@ -15,8 +67,77 @@ const Onboarding: NextPage = () => {
...
+        />
+        <TextField
```
timmyichen, line 83: placeholders i think would be helpful here, e.g. so users don't also include the `@`

pages/onboarding/index.tsx
```diff
@@ -15,8 +67,77 @@ const Onboarding: NextPage = () => {
...
+          value={formData.website}
+          onChange={handleChange}
+          fullWidth
+          inputProps={{ maxLength: 200 }}
+          required
+        />
+
+        <Box mt={4} textAlign="right">
+          <Button variant="contained" color="primary" type="submit">
+            Submit
```
timmyichen, line 122: this button should be disabled if there is an in-flight request - it will prevent them from double-clicking it accidentally.

pages/onboarding/index.tsx
```diff
@@ -1,8 +1,60 @@
...
+  linkedin_url: string;
+  github_username: string;
+  website: string;
+}
+
+interface UseridData {
+  user: UserProfile;
+}
+
+const OnboardingPage: NextPage<UseridData> = ({ user }: UseridData) => {
```
timmyichen, line 26: per your `getInitialProps`, `user` here could be `null`, so we should handle that case accordingly, otherwise the form won't work

timmyichen (changes requested):

pages/onboarding/index.tsx
jmoldyvan: makes sense, I think by that logic none should be required?

pages/onboarding/index.tsx
jmoldyvan: ok, lmk what you think of the placeholders... didn't know what to put

pages/onboarding/index.tsx
jmoldyvan: should we re-route the user to another page or set up an error?

timmyichen: i would redirect to the home page, since we don't have a dedicated login page

pages/onboarding/index.tsx
timmyichen: placeholders should guide what the users should type in. i would make some (obviously) fake examples.  e.g. for linkedin we want the full URL so maybe something like `https://www.linkedin.com/in/your-profile`

pages/onboarding/index.tsx
timmyichen: yup - though we should ensure that at least _one_ field is entered before submitting.  i.e. the submit button should be disabled until they've typed in something for at least one field

jmoldyvan: is that for any specific reason?

timmyichen: just better UX - no point in allowing them to submit zero information

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,129 @@
...
+          value={formData.website}
+          onChange={handleChange}
+          fullWidth
+          inputProps={{ maxLength: 200 }}
+        />
+        <Box mt={4} textAlign="right">
+          <Button variant="contained" color="primary" type="submit" disabled={isSubmitDisabled}>
+            {isLoading ? 'Submitting...' : 'Submit'}
+            Submit
+          </Button>
```
MainlyColors, line 122: The additional `Submit` here isn't needed , it causes `submitsubmit` to render

image

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,129 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [isFormFilled, setIsFormFilled] = useState<boolean>(false);
+
+  const router = useRouter();
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = event.target;
+    setFormData(prevFormData => ({ ...prevFormData, [name]: value }));
+    setIsFormFilled(true);
+  };
```
MainlyColors, line 43: The way this is implemented means that anytime the user changes the form the state is changed to `true` for setIsFormFilled even if the user removes the text they typed in. The button still becomes enabled so they can send another patch request with the same data they had before even if that data is exact same.

in this example, on the first `/onboarding` load after login, I filled out `bio, twitter username, linkedin url` and then came back to the page and filled out one of the empty fields partially then removed what I typed in. but I was still able to submit the data even though nothing changed
image

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,129 @@
...
+      router.push('/directory');
+    } catch (error) {
+      setIsLoading(false);
+      console.error(error);
+    }
+  };
+
+  if (!user) {
+    router.push('/');
+  }
```
MainlyColors, line 61: if there was no user then the component would never get to this bit of code, it would fail at the initial `formData` state setup 
image

cypress/e2e/base.spec.ts
```diff
@@ -20,4 +20,27 @@ describe('Visit directory page and verify users exists', () => {
...
+      cy.get('input[name="linkedin_url"]').type('https://www.linkedin.com/in/test_linkedin');
+      cy.get('input[name="github_username"]').type('test_github');
+      cy.get('input[name="website"]').type('https://www.testwebsite.com');
+      cy.visit('http://localhost:3000/directory');
+    });
+
+    cy.url().should('equal', 'http://localhost:3000/directory');
+  });
+});
+
```
MainlyColors, line 45: I think you need to login first within the test to run this test correctly or intercept the request to add login data, I'm getting an error on the Cypress test since it visits `http://localhost:3000/onboarding` right away 

image

MainlyColors (commented):
Just some things I found

pages/onboarding/index.tsx
jmoldyvan: got it, im hoping my new patch fixes it

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,129 @@
...
+const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
+  const [formData, setFormData] = useState<FormData>({
+    bio: user.bio,
+    twitter_username: user.twitter_username,
+    linkedin_url: user.linkedin_url,
+    github_username: user.github_username,
+    website: user.website,
+  });
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [isFormFilled, setIsFormFilled] = useState<boolean>(false);
```
timmyichen, line 35: nit: `<boolean>` can be removed since it can be inferred from the initial value (`false`)

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,129 @@
...
+
+  const router = useRouter();
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = event.target;
+    setFormData(prevFormData => ({ ...prevFormData, [name]: value }));
+    setIsFormFilled(true);
```
timmyichen, line 42: if they type something in, then delete it, this will remain true.  `isFormFilled` can instead just be a map over all the values of the `formData` object:

```ts
const isFormFilled = Object.values(formData).every((value) => !!value);
```

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,129 @@
...
+  };
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    try {
+      setIsLoading(true);
+      await axios.patch(`/api/users/${user.id}`, formData);
+      setIsLoading(false);
```
timmyichen, line 51: if there's an error for whatever reason, this code path won't get hit - will need to wrap the network call in a try/catch and `setIsLoading(false)` regardless of the outcome

timmyichen: Nevermind, I missed this!

timmyichen:
@jmoldyvan how's this PR going? anything you need help with / blocked by anything?

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,131 @@
...
+
+  const isSubmitDisabled = !isFormFilled && !isLoading;
 
-const Onboarding: NextPage = () => {
   return (
     <Container maxWidth="lg" className="flex justify-between pt-4">
       <Typography variant="h1" className="text-3xl">
         Onboarding
       </Typography>
+      <form onSubmit={handleSubmit}>
```
devonzara, line 73: Can we move the form out of the header of the page? See `pages/profile/[id].tsx` for reference.

cypress/e2e/base.spec.ts
```diff
@@ -20,4 +20,31 @@ describe('Visit directory page and verify users exists', () => {
...
+    cy.visit('http://localhost:3000');
+    cy.login();
+    cy.visit('http://localhost:3000/onboarding');
+    cy.get('form').within(() => {
+      cy.get('textarea[name="bio"]').type('Lorem ipsum dolor sit amet.');
+      cy.get('input[name="twitter_username"]').type('test_twitter');
+      cy.get('input[name="linkedin_url"]').type('https://www.linkedin.com/in/test_linkedin');
+      cy.get('input[name="github_username"]').type('test_github');
+      cy.get('input[name="website"]').type('https://www.testwebsite.com');
+      cy.visit('http://localhost:3000/directory');
```
devonzara, line 43: This doesn't test that the submission works... it should `click()` the submit button, the redirect will be automatic if the form submission is a success.

Also, there's nothing testing the disabled/enabled state of the login button or the presence of the loading animation. Not sure if those are things we wish to test or if that's a bit too granular.

cypress/e2e/base.spec.ts
```diff
@@ -20,4 +20,31 @@ describe('Visit directory page and verify users exists', () => {
   });
 });
 
+describe('Onboarding Page', () => {
+  afterEach(() => {
+    cy.truncateDatabase();
+  });
+
+  it('displays the correct heading', () => {
```
devonzara, line 28: This test will fail once we make `/onboarding` only accessible to logged-in users. Might want to add a `cy.login();` before the `visit()`.

That said, this isn't really testing much... do we even need it?

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,131 @@
...
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    try {
+      setIsLoading(true);
+      await axios.patch(`/api/users/${user?.id}`, formData);
+      setIsLoading(false);
+      router.push('/directory');
+    } catch (error) {
+      setIsLoading(false);
+      console.error(error);
```
devonzara, line 62: We shouldn't really have `console` output on the client... if there may be an error here, we should probably show something on the screen for the user.

cypress/e2e/base.spec.ts
```diff
@@ -20,4 +20,31 @@ describe('Visit directory page and verify users exists', () => {
...
+    cy.visit('http://localhost:3000/onboarding');
+    cy.contains('h1', 'Onboarding');
+  });
+
+  it('fills out the form successfully', () => {
+    cy.visit('http://localhost:3000');
```
devonzara, line 34: This first `visit()` isn't necessary.

devonzara (changes requested):
Few minor things... otherwise, looking good to me!

pages/onboarding/index.tsx
jmoldyvan: is there a benefit to doing this?

cypress/e2e/base.spec.ts
jmoldyvan: I mean, idk what the threshold for testing really is

pages/onboarding/index.tsx
```diff
@@ -59,7 +59,7 @@ const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
       router.push('/directory');
     } catch (error) {
       setIsLoading(false);
-      console.error(error);
+      alert(error);
```
devonzara, line 62: `alert()` doesn't create a very good user experience and should be avoided in most cases.

Since we're using MUI, a toast alert is pretty easy to do. It seems that MUI calls them [Snackbars](https://mui.com/material-ui/react-snackbar/).

cypress/e2e/onboarding.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  it('displays the correct heading', () => {
+    cy.login();
+    cy.visit('http://localhost:3000/onboarding');
+    cy.contains('h1', 'Onboarding');
+  });
+
+  it('fills out the form successfully', () => {
+    cy.get('form').within(() => {
+      cy.get('textarea[name="bio"]').type('Lorem ipsum dolor sit amet.');
```
devonzara, line 16: Too much was removed here, thus causing the tests to fail.

You had:
```js
    cy.visit('http://localhost:3000');
    cy.login();
    cy.visit('http://localhost:3000/onboarding');
    [...]
```
Only the first `cy.visit()` should have been removed, so:
```js
    cy.login();
    cy.visit('http://localhost:3000/onboarding');
    [...]
```

devonzara (changes requested):

pages/onboarding/index.tsx
devonzara: A form like this shouldn't be in the header... making this change now will simplify the tasks later when we add proper page layouts to the application.

cypress/e2e/onboarding.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+describe('Onboarding Page', () => {
+  afterEach(() => {
+    cy.truncateDatabase();
+  });
+
+  it('displays the correct heading', () => {
+    cy.login();
+    cy.visit('http://localhost:3000/onboarding');
+    cy.contains('h1', 'Onboarding');
+  });
```
timmyichen, line 12: we can probably do without this test haha

cypress/e2e/onboarding.spec.ts
```diff
@@ -0,0 +1,28 @@
+/// <reference types="cypress" />
+
+describe('Onboarding Page', () => {
+  afterEach(() => {
+    cy.truncateDatabase();
+  });
```
timmyichen, line 6: is this necessary to do for onboarding related tests?

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,131 @@
...
+      await axios.patch(`/api/users/${user?.id}`, formData);
+      setIsLoading(false);
+      router.push('/directory');
+    } catch (error) {
+      setIsLoading(false);
+      alert(error);
+    }
+  };
+
+  const isSubmitDisabled = !isFormFilled && !isLoading;
```
timmyichen, line 66: i'm not sure if this logic is what we want - i'm reading this as: "submit is disabled if the form is not filled, and it is not loading" - meaning:
* if `isLoading` = true, then `isSubmitDisabled` = false

we can break this down into thinking: when do we want the button to be disabled?
* when the form is not filled (nothing to submit)
* when they're in the middle of a submission (avoid double submits)

these two criteria can be respectively expressed as:
* `!isFormFilled` (which you have), OR
* `isLoading`

```suggestion
  const disableSubmit = !isFormFilled || isLoading;
```

pages/onboarding/index.tsx
timmyichen: i also don't think we need a heading.  i.e. we don't need a title for the page that says "onboarding"... onboarding is mostly an internal concept.  we should be communicating this to our users in a more user-friendly way, like "Complete your profile to connect with other members!"

pages/onboarding/index.tsx
```diff
@@ -1,15 +1,131 @@
...
+          placeholder='https://www.linkedin.com/in/your-profile'
+          value={formData.linkedin_url}
+          onChange={handleChange}
+          fullWidth
+          inputProps={{ maxLength: 200 }}
+        />
+        <TextField
+          label="GitHub Username"
+          name="github_username"
+          placeholder='Jon Doe'
```
timmyichen, line 106: this technically isn't a valid username, also i think we should avoid gendered examples - would change this (plus the placeholder twitter handle above) to be more generic, e.g. `twitteruser` and `githubuser`

cypress/e2e/onboarding.spec.ts
devonzara: I'd argue it doesn't hurt to clean up... get rid of any users created so there's a fresh slate for whatever tests may run after this file. Granted, it could be an `after()`...

pages/onboarding/index.tsx
```diff
@@ -59,16 +68,23 @@ const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
       router.push('/directory');
     } catch (error) {
       setIsLoading(false);
-      alert(error);
+      setOpenAlertError(error);
```
devonzara, line 71: > Error: Objects are not valid as a React child (found: object with keys {message, name, code, config, request, response}). If you meant to render a collection of children, use an array instead.

`setOpenAlertError()` expects a `ReactChild`, not an object. A string would be a valid `ReactChild` so it should be `error.message`...

That said, we could probably just put a generic error in here... doing so would be less confusing to non-technical users.

pages/onboarding/index.tsx
```diff
@@ -131,6 +147,11 @@ const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
           </Button>
         </Link>
       </Box>
+      <Snackbar open={openAlertError} autoHideDuration={6000} onClose={handleClose}>
```
devonzara, line 150: nit: I'd probably opt to have them in the bottom-right rather than the bottom-left. Personal opinion though...

devonzara (changes requested):

pages/onboarding/index.tsx
MainlyColors: This is redirecting now but since it was done in a useEffect, the page loads completely and flashes at the user before redirecting.

pages/onboarding/index.tsx
MainlyColors: This still doesn't solve the issue because the state is queued so its always on a 1 character delay with this setup.

example, I have a empty form and write "test" in bio then backspace it

based on this code 
```ts
  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    const { name, value } = event.target;
    setFormData(prevFormData => ({ ...prevFormData, [name]: value }));
    console.log(formData);
    const isFormFilled = Object.values(formData).some(value => !!value);
    console.log(isFormFilled);
    setIsFormFilled(isFormFilled);
  };
```
```
index.tsx:55 {bio: '', twitter_username: '', linkedin_url: '', github_username: '', website: ''}
index.tsx:57 false
index.tsx:55 {bio: 't', twitter_username: '', linkedin_url: '', github_username: '', website: ''}
index.tsx:57 true
index.tsx:55 {bio: 'te', twitter_username: '', linkedin_url: '', github_username: '', website: ''}
index.tsx:57 true
index.tsx:55 {bio: 'tes', twitter_username: '', linkedin_url: '', github_username: '', website: ''}
index.tsx:57 true
index.tsx:55 {bio: 'test', twitter_username: '', linkedin_url: '', github_username: '', website: ''}
index.tsx:57 true
index.tsx:55 {bio: 'tes', twitter_username: '', linkedin_url: '', github_username: '', website: ''}
index.tsx:57 true
index.tsx:55 {bio: 'te', twitter_username: '', linkedin_url: '', github_username: '', website: ''}
index.tsx:57 true
index.tsx:55 {bio: 't', twitter_username: '', linkedin_url: '', github_username: '', website: ''}
index.tsx:57 true
```

at the end since there is a 1 character delay, I can still submit

image

pages/onboarding/index.tsx
jmoldyvan: yeah idk i liked it in the left. i guess we can consult design team?

devonzara: Fair, I believe my bias may come from operating systems as they tend to default to top-right or bottom-right.

Anyway, this is non-blocking and can always be changed later easily enough. 👍🏼

pages/onboarding/index.tsx
devonzara: @jmoldyvan Tim merged #95 which gives us proper layout components. When you merge changes from `main`, this file can be simplified.

devonzara (changes requested):
Few changes suggested... Will also want to re-review after you merge `main` into this PR.

pages/onboarding/index.tsx
```diff
@@ -68,7 +68,7 @@ const OnboardingPage: NextPage<Props> = ({ user }: Props) => {
       router.push('/directory');
     } catch (error) {
       setIsLoading(false);
-      setOpenAlertError(error);
+      setOpenAlertError(true);
     }
```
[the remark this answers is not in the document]
jmoldyvan: what do you mean? like we should be specializing our error catches instead of just in general firing on all server errors?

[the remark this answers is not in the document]
timmyichen: @devonzara not sure what you mean - validation errors in terms of validating the client input fields before sending the request off to the server?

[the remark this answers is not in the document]
devonzara: @jmoldyvan Sorta. Any input validation errors, specifically, should be handled separately.

@timmyichen Before is good too... but more importantly, when it hits the server. If input validation on the server fails, that should be reflected in the form so the user knows what to change and why.

Currently, if you remove the `maxlength="1000"` from the bio input field and type 1001 characters, you'll get this error - which is probably not an error we want to be sending back to the client.

image
```html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Error<br> &nbsp; &nbsp;at Query.run (/home/user/workbench/dev-directory/node_modules/sequelize/src/dialects/postgres/query.js:76:25)<br> &nbsp; &nbsp;at /home/user/workbench/dev-directory/node_modules/sequelize/src/sequelize.js:650:28<br> &nbsp; &nbsp;at processTicksAndRejections (node:internal/process/task_queues:95:5)<br> &nbsp; &nbsp;at async PostgresQueryInterface.bulkUpdate (/home/user/workbench/dev-directory/node_modules/sequelize/src/dialects/abstract/query-interface.js:930:12)<br> &nbsp; &nbsp;at async Function.update (/home/user/workbench/dev-directory/node_modules/sequelize/src/model.js:3280:28)</pre>
</body>
</html>
```
In other words, I believe the toast errors should be used to show non-validation errors and validation errors should appear inline next to their respective fields.

Though, after seeing that there isn't really any server-side validation... I suppose this ought to be a separate ticket as that extends the scope of this quite a bit.

timmyichen:
@jmoldyvan any updates?

timmyichen (approved):

Caleb-Cohen (approved):
LGTM!

PR #88, opened 2023-04-17
discord_name should never be empty since we use discord auth

MainlyColors:
# Description

Found a small bug in the implementation of `makeUserObject` which `discord_name` has a random empty chance. There should be no scenario where we don't get the `discord_name` since we are using discord auth to set up users

## Testing
not much to test, one line change

## Type of change

# Checklist:

- [ ] I have added comments on any new, hard-to-understand code

### DIFF
--- a/server/test/utils.ts
+++ b/server/test/utils.ts
@@ -43,7 +43,7 @@ function makeUserObject({
   return {
     email: email ?? randEmail(),
     discord_user_id: discord_user_id ?? String(randNumber({ min: 1e16, max: 1e18 - 1 })),
-    discord_name: discord_name ?? randomEmptyChance(20, getRandomDiscordUserName()),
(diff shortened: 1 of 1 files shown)


timmyichen (approved):

MainlyColors:
> This is one example of why I typically use nullable fields... if we were doing a random chance that the fields be `NULL`, this would have been throwing errors that we could have caught sooner.
> 

The purpose of the empty string instead of NULL was to mimic if a new user didn't fill out a certain field for any reason. 

I'll look into what Random chance for NULL will do in different parts of the app and the Sequelize link.

devonzara:
@MainlyColors Sorry, I wasn't suggesting that we switch to `NULL` fields now... I was merely mentioning it because it was something I had brought up in a voice chat last week.

That said, the `notEmpty` validator might be worth adding.

PR #89, opened 2023-04-18
#78 create flags that check if user has been through onboarding

### LINKED ISSUE #78: Create Flags That Check If User Has Been Through Onboarding

DevinCLane:
Closes #78 

# Description

Pretty bigly PR.  Does a bunch of things:

Backend stuff
* Adds a `flags` table to allow setting of flags on users.  Used in this case to flag whether a user has skipped onboarding
* Adds new endpoints:
  * `GET /api/auth/current-user` which returns the logged-in user and flags
  * `PUT /api/user/onboarding/skip` which upserts the skipped onboarding flag for the logged-in user
* Modifies endpoints:
  *  `POST /api/auth/login` returns flags as well
* Adds a getter `profile` to the User model that will get us the public, client-friendly object, so we don't have to call `_.pick` every time we want to send that data along

Testing:
* Adds/updates tests so new/updated endpoints are all passing
* Removed the `truncateDatabase` function as this is an antipattern and we should generally not allow a single test to impact other tests
* Thanks to the above change, tests can now run in parallel
* Added the ability for tests to run on a specific, fresh database, so that _if_ the global DB state matters, it's easier to manage/predict
* Cypress tests now have a baseURL so you don't have to include `localhost:3000` in all of the `cy.visits`

Frontend stuff:
* Adds flags to the auth store
* App uses the new current-user route for getting logged-in user data, plus populating the context instead of the public profile/id one.
* 

## Testing
Manually by clearing the DB, logging in, clicking "skip onboarding", then logging out, logging in again, and it should put you right on `/directory` and not the onboarding page

Plus server tests
image

## Type of change

## Screenshots

Please include any screenshots if there are visual changes.

# Checklist:


server/api/users/users.controller.ts
```diff
@@ -92,10 +92,32 @@ const updateUserById: RequestHandler<{ id: string }, string, UpdatableFields> =
   res.sendStatus(200);
 };
 
+interface FlagAttributes {
+  id: number;
+  user_id: number;
+  flag_name: string;
+}
+
+const getUserFlags: RequestHandler<{ id: string }, { message: string } | { flags: string[] }> = async (req, res) => {
```
timmyichen, line 101: this gets _any_ user's flags, but i think we only ever want to expose a user's own flags - so this route should probably not have an id parameter, and only return the flags of the currently-logged-in-user.

server/api/users/users.test.ts
```diff
@@ -207,4 +209,38 @@ describe('user router', () => {
...
   });
+
+  describe('Create new flag', () => {
+    it('should create a new flag and verify the data', async () => {
+      const newFlag = {
+        name: 'test flag',
+        description: 'this is a test flag',
+        isActive: true,
+      };
+      const response = await axios.post('/api/flags', newFlag);
```
timmyichen, line 220: does this route exist? i don't think we should have an endpoint that allows arbitrary flag creation - it should be managed solely on the server side

server/migrations/20230412233941-create-flags-table.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable('flags', {
+      id: {
+        type: Sequelize.INTEGER,
+        autoIncrement: true,
+        primaryKey: true,
```
timmyichen, line 10: i'd also make this non-nullable

timmyichen (commented):
i think this is the right direction! is there anything in particular you feel like you are struggling with?

server/models/Flag.model.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+Flag.init(
+  {
+    id: {
+      type: DataTypes.INTEGER,
+      autoIncrement: true,
+      primaryKey: true,
+    },
+    user_id: {
```
timmyichen, line 27: we should ensure this is a FK to users as well

server/api/users/users.test.ts
```diff
@@ -207,4 +207,29 @@ describe('user router', () => {
       });
     });
   });
+
+  describe('Flag API', () => {
```
timmyichen, line 211: these tests seem to be testing the model, rather than the API 👀

server/models/User.model.ts
```diff
@@ -28,6 +30,28 @@ class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {
...
     ];
   }
+
+  public static associations: {
+    flags: Association<User, Flag>;
+  };
+
+  public static associate() {
+    User.hasMany(Flag, { foreignKey: 'user_id' });
+  }
```
MainlyColors, line 40: The docs mention we also need a `Flag.belongsTo(User)` otherwise only the `User` model knows about the association while the `Flag` model does not

based on this example
```js
Team.hasMany(Player);
Player.belongsTo(Team);
```

I'm confused about the example where this is supposed to go though but I think after `User.init` step because its shown like that in another [example](https://sequelize.org/docs/v6/advanced-association-concepts/creating-with-associations/#belongsto--hasmany--hasone-association)



https://sequelize.org/docs/v6/core-concepts/assocs/#implementation-1
https://sequelize.org/docs/v6/core-concepts/assocs/#why-associations-are-defined-in-pairs

server/models/User.model.ts
```diff
@@ -85,4 +109,8 @@ User.init({
   },
 });
 
+User.associations = {
+  flags: User.hasMany(Flag, { foreignKey: 'user_id' }),
+};
+
```
MainlyColors, line 115: What is this bit of code doing? Isn't the `hasMany` association already defined within the static `associate` method?

I couldn't find in the docs where associations are defined like this

MainlyColors (commented):
I'm going to skip commenting on the tests since They test Model and not the API so I expect they will change. See my other comments about the models

server/models/User.model.ts
MainlyColors: Based on voice chat, Devon's & Tim's comments were as it stands we don't need to be able to search users from the Flag models so a `Flag.belongsTo(Users)` isn't nesccarry

timmyichen:
@DevinCLane do you need re-review, or is there anything we can do to help with this PR?

DevinCLane:
I'm not feeling well, so if this is blocking feel free to have someone else take it over! A re-review could be good in the meantime?

docker-compose.ci.yml
```diff
@@ -10,7 +10,6 @@ services:
       DB_HOST: ${DB_HOST:-pg}
       DB_PORT: ${DB_PORT:-5432}
       DB_USER: ${DB_USER:-postgres}
-      DB_NAME: ${DB_NAME:-dev-directory}
```
timmyichen, line 13: this environment variable was overriding the one set by `.env.test`, hence the removal

cypress/e2e/dev.spec.ts
```diff
@@ -34,15 +30,15 @@ describe('example run-util test', () => {
...
 //     cy.get('ul li').should('have.length', 0);
 //
 //     cy.createUsers(20);
-//     cy.visit('http://localhost:3000/directory');
+//     cy.visit('/directory');
 //     cy.get('ul li').should('have.length', 20);
 //
 //     cy.truncateDatabase();
-//     cy.visit('http://localhost:3000/directory');
+//     cy.visit('/directory');
```
Caleb-Cohen, line 41: Do we still need this if it's commented out?

Caleb-Cohen (approved):
This is such a massive PR, but it does LGTM.

cypress/e2e/dev.spec.ts
timmyichen: @devonzara added this with the comment:

> // Example test would be too closely tied to the temporary UI
> // Left here for reference, but should remain commented.

Leaving it in for him to decide whether it should stay or not

windyScripts (approved):
The refactoring and improvements are quite impressive.

Tests ran fine for me:
image

PR #90, opened 2023-04-19
#60 implement infinite scroll

### LINKED ISSUE #60: Implement Infinite Scroll
**Related Issues**
- Project: #17  
- Depends on: 
  - #58 
  - #59   

**Description**
- Use [Intersection Observer](https://www.freecodecamp.org/news/how-to-implement-infinite-scroll-in-next-js/) to trigger loading of a new batch of 20 profiles on the home page when the user scrolls down.
- Ensure we don't load pages that don't exist (i.e., if the server tells us we have 5 total pages, we never request a 6th page.)
- The user should see that the next batch of pages is loading (i.e., there is a loading state visualized).
- A page is never loaded twice. 
- When loading the next page, we are not fetching an extra batch of pages (i.e., if the user triggers a new load, then scrolls up and back down before the loading completes, we do not trigger loading a new batch of pages).

**Acceptance Criteria**
- A new batch of 20 profiles is displayed when the user scrolls down.
- We don't load pages that don't exist.
- The user sees a loading state when a new batch of pages is loading.
- A page is never loaded twice.
- One batch of pages is loaded at a time.
- Acceptance tests exist for this issue.

MainlyColors:
Closes #60

# Description

## feature
- Use `Intersection Observer` to trigger loading of a new batch of 20 profiles on the `/directory` page when the user scrolls down to the height of the last card above the bottom of the window/ul that contains all the users
 -  Will not load pages past the total pages returned by `/api/users`
 - only triggers the load event once because the observer is removed on first intersection
 - A loading spinner is triggered if the data doesn't load instantly
 
 ## Tests added
 - cypress tests added for all functionality
 - seed script is triggered before all tests in the `directory infinite scroll` test suite to generate 100 users

#### manually
- start docker container
- `npm run dev:server`
- `npm run seed` - to generate a 100 users - 
- visit `\directory` 
  - keep in mind currently `discord_name` has a 20% chance of being empty which is being fixed in another PR but will cause some users not to show up on the page
- scroll down the page to load a new 20 users (repeat until no more users)
- in the network tab can change the throttle to `slow 3G` to catch the loading spinner 

#### Automatic
- start docker container
- `npm run dev:server`
- `npm run cypress:open`
- click "Start E2E Test" with Chrome
- Tests will run and should all pass

## Type of change

## Screenshots

Tests passing - directory.spec.ts
image

## Known issue - intercepting user api w/ cypress

The use of cypress intercept here in the directory test is intended to wait upon the response and catch the loading spinner. This works 95-99% of the time but if the loading spinner appears for less than 20ms then cypress won't be able to catch it, resulting in a failed test. Just re-run the test to be sure.

Different variations of using `as()`, `(req) => req.alias`, and even `req.reply(res => res.setDelay(100))` and every possible permutation of those 3 were tried by Myself and Devon but we still got random errors.

```ts
it('shows a loading animation while fetching users', () => {
    cy.scrollTo('bottom');
    cy.intercept('/api/users?page=*').as('getUsers');

    // spinner should exist & be removed after request completes
    cy.get('[data-cy="loading"]');
    cy.wait('@getUsers');
    cy.get('[data-cy="loading"]').should('not.exist');
  });
```

# Checklist:


cypress/e2e/base.spec.ts
```diff
@@ -20,4 +20,94 @@ describe('Visit directory page and verify users exists', () => {
...
+      // To make this code work you need to setup 2 cypress env variables in cypress.config.ts
+      // - bash (pointing to the bash shell executable on your machine)
+      // - comSpec (pointing to cmd.exe on your machine)
+      const sh = `"${Cypress.env('bash')}" --login -c`;
+      const fullCommand = `${sh} "${command}"`;
+      // const sh = `"${Cypress.env('bash')}" --login -i -c`;
+      cy.exec(fullCommand, { env: { SHELL: Cypress.env('comSpec') } })
+        .its('code')
+        .should('eq', 0);
+    }
```
MainlyColors, line 80: This runs the seed script to generate 100 users. But I haven't found a way to remove the users from the DB within cypress. you can see attempts I made to do this above this code.

pages/directory/index.tsx
```diff
@@ -8,6 +9,56 @@ import createAxiosInstance from 'client/lib/axios';
...
+      try {
+        // stop loading users when at max pages
+        if (currentPage + 1 > totalPages) {
+          setIsLoading(false);
+          return;
+        }
+        const { page, total, users } = await fetchUsers(currentPage + 1);
+        setCurrentPage(page);
+        setTotalPages(total);
+        setUserData([...userData, ...users]);
```
MainlyColors, line 53: using `totalPages` state here to avoid an additional api call, I understand this could be handled by calling the api and if a 400 badRequest happens then we use that to determine the max pages that were hit. But this way after the first request we know the totalPages to look for and when those are hit then we can prevent any additional api calls

pages/directory/index.tsx
```diff
@@ -8,6 +9,56 @@ import createAxiosInstance from 'client/lib/axios';
...
+    }
+  }
+
+  React.useEffect(() => {
+    const intersectionOptions = {
+      // root default = document's viewport
+      // in the future offset should be tied to the card height of the last row
+      rootMargin: '300px',
+      threshold: 0,
+    };
```
MainlyColors, line 37: 300px is a temp value, In the future, the offset should be based on the height of one row of user profiles

cypress.config.ts
```diff
@@ -10,4 +10,8 @@ export default defineConfig({
   videosFolder: './cypress/videos',
   downloadsFolder: './cypress/downloads',
   fileServerFolder: './',
+  env: {
+    bash: 'C:\\Program Files\\Git\\bin\\bash.exe',
+    comSpec: 'C:\\Windows\\system32\\cmd.exe',
+  },
```
MainlyColors, line 16: required for the `cy.exec()` work around on windows machines

cypress/e2e/base.spec.ts
```diff
@@ -20,4 +20,94 @@ describe('Visit directory page and verify users exists', () => {
   });
 });
 
+describe('directory infinite scroll', () => {
+  const PAGE_LIMIT = 20;
```
MainlyColors, line 24: I know we use this here `server\api\users\users.controller.ts` too so should probably be abstracted away to an env property or something?

Is env the best way to handle app settings for something like `page_limit`?

cypress/e2e/base.spec.ts
```diff
@@ -20,4 +20,94 @@ describe('Visit directory page and verify users exists', () => {
...
+
+  it('first 20 users load ', () => {
+    cy.get('[data-cy="user-container"]').children().should('have.length', 20);
+  });
+
+  it('Should show the loading spinner when loading the data then remove it afterwards', () => {
+    cy.scrollTo('bottom');
+    cy.intercept('GET', '/api/users?page=*', req => {
+      req.on('response', res => res.setDelay(500));
+    }).as('getUsers');
```
MainlyColors, line 97: intercepts the api call to add a 500 ms delay to give cypress enough time to see the loading spinner

cypress/e2e/base.spec.ts
```diff
@@ -20,4 +20,94 @@ describe('Visit directory page and verify users exists', () => {
...
+    });
+
+    // clear the db then fill with 100 users
+    // User.destroy({ truncate: true })
+    //   .then(() => {
+    //     runSeedCommand();
+    //   })
+    //   .catch(err => {
+    //     console.error('Error deleting existing users:', err);
+    //   });
```
MainlyColors, line 60: I couldn't get Cypress to accept any imports so this errors. But the intention is I believe we should clear the DB before seeding the DB with 100 users so we know what to expect.

MainlyColors:
I realized that I forgot to test under 20 users, so I need to write a test for that and fix the code because under 20 users leaves a forever loading state showing

pages/directory/index.tsx
```diff
@@ -1,13 +1,64 @@
...
```
devonzara, line 37: Instead of using a magic number here, could we select a card and calculate its height? Something along the lines of:

```js
const height = getElementHeight(document.querySelector('card selector'));

function getElementHeight(el) {
  const rect = el.getBoundingClientRect();
  const { marginTop, marginBottom } = getComputedStyle(el);

  return rect.height + parseFloat(marginTop) + parseFloat(marginBottom);
}
```

server/scripts/resetDB.ts
devonzara, line 1: This should no longer be needed after #79 as it implements [a function that truncates all tables](https://github.com/timmyichen/dev-directory/blob/372e31a10126d04c4dbe856afefd67e373cd9b49/server/test/utils.ts#L74)... this also has the benefit of resetting the auto-incrementing user `id`.

devonzara (changes requested):
Not sure about the stance of others on using `cy.exec()`... Is it the recommended approach? What does Cypress recommend using it for?

#79 implements a way to seed and reset the database, as well as run other test utilities. If we decide we don't need `cy.exec()`, using it would simplify this PR quite a bit.

Otherwise, looks good, just a few nits.

timmyichen:
@MainlyColors anything we can do to help with this PR?

MainlyColors:
@timmyichen Nothing as of right now, I just got back from vacation so ill be going through all the comments tomorrow (5/8) and making the changes then

cypress/support/e2e.ts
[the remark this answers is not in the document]
MainlyColors: @devonzara Cypress references a `resetDB` function in their best practices documentation. It's a custom Command they made for a specific project that uses a local JSON database so I tried to apply the same ideas with our SQL database use case 

https://docs.cypress.io/guides/references/best-practices#Its-all-downside-with-no-upside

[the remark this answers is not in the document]
MainlyColors: Ill swap out what I have here for the #91 PR approach

cypress.config.ts
MainlyColors: This is only a Windows issue, so this bit of code checks the platform and if its not `win32` then run `cy.exec()` like normal. The bug comes from an issue on Cypress's side where they insert characters into windows paths leading to cypress not being able to find the command line bin.
https://github.com/timmyichen/dev-directory/blob/47e962d85acb025769c1662c74ba0cee5b4830c0/cypress/support/e2e.ts#L19-L23

cypress/e2e/directory.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+      cy.intercept('GET', '/api/users?page=*', req => {
+        req.on('response', res => res.setDelay(500));
+      }).as('getUsers');
+      cy.scrollTo('bottom');
+
+      // last page won't have a spinner
+      if (i === totalPages) break;
+      // check spinner happened & removed
+      cy.get('[data-cy="loading"]'); //check if exists
+      cy.get('[data-cy="loading"]').should('not.exist'); // removed after data loaded
```
devonzara, line 42: nit: formatting, empty line after breaks/continues & same line comments should typically be avoided

```suggestion
      // last page won't have a spinner
      if (i === totalPages) break;

      // spinner should exist & be removed after request completes
      cy.get('[data-cy="loading"]');
      cy.get('[data-cy="loading"]').should('not.exist');
```  
Also, is this missing a `cy.wait` or something? How does it know when the request is complete?

cypress/e2e/directory.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  it('shows a loading animation while fetching users', () => {
+    cy.scrollTo('bottom');
+    cy.intercept('GET', '/api/users?page=*', req => {
+      req.on('response', res => res.setDelay(500));
+    }).as('getUsers');
+
+    cy.get('[data-cy="loading"]'); //check if exists
+    cy.wait('@getUsers');
+    cy.get('[data-cy="loading"]').should('not.exist'); // removed after data loaded
```
devonzara, line 58: nit: same as above, inline comments

pages/directory/index.tsx
```diff
@@ -1,13 +1,72 @@
...
+  async function fetchUsers(page: number) {
+    try {
+      const response = await axios.get(`/api/users?page=${page}`);
+      return {
+        page: response.data.page,
+        total: response.data.totalPages,
+        users: response.data.users,
+      };
+    } catch (err) {
+      console.error(err);
```
devonzara, line 27: We should avoid using `console` on the client side as it will fail without notifying the user... see linked comments on #87 and the approach they took to address it.

https://github.com/timmyichen/dev-directory/pull/87#discussion_r1185713660
https://github.com/timmyichen/dev-directory/pull/87#discussion_r1186645305

pages/directory/index.tsx
```diff
@@ -1,13 +1,72 @@
...
+
+    const intersectionOptions = {
+      // root default = document's viewport
+      rootMargin: `${height}px`,
+      threshold: 0,
+    };
+
+    const lastCardObserver = new IntersectionObserver(async ([lastCard]) => {
+      // guard clause, should intersect only once
+      if (!lastCard.isIntersecting) return;
```
devonzara, line 49: nit: add linebreak after `return;` to make it more readable by adding a separation in logic.

pages/directory/index.tsx
```diff
@@ -1,13 +1,72 @@
...
+        if (currentPage + 1 > totalPageNum) {
+          setIsLoading(false);
+          return;
+        }
+        const { page, total, users } = await fetchUsers(currentPage + 1);
+        setCurrentPage(page);
+        setTotalPageNum(total);
+        setUserData([...userData, ...users]);
+      } catch (err) {
+        console.error(err);
```
devonzara, line 63: Same as above, we should avoid using `console` in the client-side code.

server/scripts/resetDB.ts
devonzara, line 1: File is no longer needed in this PR.

devonzara (changes requested):
Apologies for the delay... I've not gotten the chance to pull it down and actually work through/test the PR, but I've added some comments/suggested changes that can be done in the interim. 👍🏼

cypress/e2e/directory.spec.ts
MainlyColors: @devonzara Cypress will keep retrying their commands for 4 secs each before it declares it failed. That's why a `cy.wait` isn't required

cypress/e2e/directory.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    cy.visit('http://localhost:3000/directory');
+  });
+
+  function checkInfiniteScroll(pages: number, totalPages: number) {
+    for (let i = 1; i <= pages; i++) {
+      cy.get('[data-cy="user-container"]').children().as('ulChildren');
+      i < totalPages ?
+        cy.get('@ulChildren').should('have.length', PAGE_LIMIT * i) :
+        // last page not guaranteed to be full
+        cy.get('@ulChildren').should('have.length.within', PAGE_LIMIT * (i - 1), PAGE_LIMIT * i);
```
timmyichen, line 28: stylistic choice, but we should reserve ternary statements for values, rather than actual execution.  an if/else would be preferred here

pages/directory/index.tsx
```diff
@@ -31,11 +108,11 @@ Directory.getInitialProps = async ({ req }) => {
...
     const axios = createAxiosInstance(req);
     const response = await axios.get('/api/users');
-    const users = response.data.users;
+    const { users, totalPages } = response.data;
 
-    return { users, error: null };
+    return { users, totalPages, error: null };
   } catch (error) {
-    return { users: [], error: error.message };
+    return { users: [], totalPages: 1, error: error.message };
```
timmyichen, line 115: why is `totalPages` set to 1 here?

pages/directory/index.tsx
```diff
@@ -1,28 +1,105 @@
...
-const Directory: NextPage<{ users: ClientUser[]; error: string }> = props => {
-  const { users, error } = props;
+const Directory: NextPage<{ users: ClientUser[]; totalPages: number; error: string }> = props => {
+  const { users, totalPages, error } = props;
+  const lastCardRef = React.useRef(null);
+  const [userData, setUserData] = React.useState(users);
+  const [currentPage, setCurrentPage] = React.useState(1);
+  const [totalPageNum, setTotalPageNum] = React.useState(totalPages);
+  const [isLoading, setIsLoading] = React.useState(false);
+  const [open, setOpen] = React.useState(false);
```
timmyichen, line 17: would prefer to have this be named a little more specifically (if we have more than one alert/modal/etc, what this opens becomes ambiguous).  `showError, setShowError` as a sample alternative

pages/directory/index.tsx
```diff
@@ -1,28 +1,105 @@
...
+      if (!lastCard.isIntersecting) return;
+
+      lastCardObserver.disconnect();
+      setIsLoading(true);
+      try {
+        // stop loading users when at max pages before fetching next page
+        if (currentPage + 1 > totalPageNum) {
+          setIsLoading(false);
+          return;
+        }
```
timmyichen, line 61: can this check happen before we even set `isLoading`?  that way we can just `setIsLoading(true)` right before we start fetching the next page of users

pages/directory/index.tsx
```diff
@@ -1,28 +1,105 @@
...
+      }
+    }, intersectionOptions);
+
+    setIsLoading(false);
+    lastCardObserver.observe(lastCardRef.current);
+  }, [userData]);
 
   return (
     <Container maxWidth="lg" className="pt-4">
+      <ErrorPopup open={open} setOpen={setOpen} message="Page out of range" />
```
timmyichen, line 79: A note - `Page out of range` isn't the only possible error we could get - but I think error handling in general can be worked on later so I won't block on this :P

pages/directory/index.tsx
```diff
@@ -1,28 +1,105 @@
...
+        setTotalPageNum(total);
+        setUserData([...userData, ...users]);
+      } catch (err) {
+        if (err.message === 'Page out of range') {
+          setOpen(true);
+        }
+      }
+    }, intersectionOptions);
+
+    setIsLoading(false);
```
timmyichen, line 73: was this meant to be inside of the observer callback?

timmyichen (commented):
minor comments, nothing major, i think this is almost ready to go!

pages/directory/index.tsx
MainlyColors: Originally I put it there for a reason I can't remember, but within the observer callback makes more sense 👍

pages/directory/index.tsx
```diff
@@ -1,28 +1,106 @@
...
+  async function fetchUsers(page: number) {
+    try {
+      const response = await axios.get(`/api/users?page=${page}`);
+      return {
+        page: response.data.page,
+        total: response.data.totalPages,
+        users: response.data.users,
+      };
+    } catch (err) {
+      if (err.code === 'ERR_BAD_REQUEST') {
```
devonzara, line 29: What happens if other errors occur? Also, is `ERR_BAD_REQUEST` only going to occur when the requested page is out of range?

pages/directory/index.tsx
```diff
@@ -1,28 +1,106 @@
...
+        return;
+      }
+      setIsLoading(true);
+      try {
+        const { page, total, users } = await fetchUsers(currentPage + 1);
+        setCurrentPage(page);
+        setTotalPageNum(total);
+        setUserData([...userData, ...users]);
+      } catch (err) {
+        if (err.message === 'Page out of range') {
```
devonzara, line 67: I'm not sure this conditional is necessary... or, if kept - should we provide an `else` condition with a generic error?

devonzara (changes requested):
Looking good, mostly nits and questions about error handling remain.

pages/directory/index.tsx
MainlyColors: I didn't know all the errors that were possible, Ill look into it more

devonzara: Every possible error doesn't need to be explicitly handled; the user doesn't always need to know what the error was.

Some errors may be network related and work themselves out, others might be things we need to address... either way, communicating that an error has occurred to the user instead of leaving them uncertain results in a better user experience.

cypress/e2e/directory.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+      if (i < totalPages) {
+        cy.get('@ulChildren').should('have.length', PAGE_LIMIT * i);
+      } else {
+        // last page not guaranteed to be full
+        cy.get('@ulChildren').should('have.length.within', PAGE_LIMIT * (i - 1), PAGE_LIMIT * i);
+      }
+
+      cy.intercept('GET', '/api/users?page=*', req => {
+        req.on('response', res => res.setDelay(500));
+      }).as('getUsers');
```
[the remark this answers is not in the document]
MainlyColors: While researching about `.as()` more and if `cy.wait()` was necessary, I found the correct way to actually pause the response while waiting for the loading spinner


```ts
      cy.intercept('/api/users?page=*', req => {
        /**
       * If set, `cy.wait` can be used to await the request/response
       * cycle to complete for this request via `cy.wait('@alias')`.
       */
        req.alias = 'getUsers';
      }).as('getUsers');
      
      cy.get('[data-cy="loading"]');
      // here cypress releases the response
      cy.wait('@getUsers');
      cy.get('[data-cy="loading"]').should('not.exist');
```

client/components/ErrorPopup.tsx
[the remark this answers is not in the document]
MainlyColors: I agree on ErrorToast for now, I think configuring the component to handle situations that haven't occurred yet and may never occur goes against YAGNI. In the future, if we need the toast to do more then we can add the functionality then 😎

Caleb-Cohen (approved):
LGTM

cypress/e2e/directory.spec.ts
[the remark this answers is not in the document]
devonzara: @MainlyColors Correct me if I'm wrong, but I believe `.as()` and `req.alias =` have similar behaviors, and are therefore redundant when used together. Given that dynamic alias assignment is not a requirement in this context, using `.as()` alone should suffice. No?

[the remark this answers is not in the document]
MainlyColors: its not the same, I had `.as()` before with `.wait()` and its not doing the same thing.

`req.alias` is telling Cypress to look for when the alias shows up to release the response. Cypress is holding back the response and then when it reads wait for the same alias, it releases the response.

devonzara (approved):
LGTM. Thanks for taking this on!

PR #91, opened 2023-04-28
🧪 test(cypress): add dev-only routes & commands for testing

devonzara:
Closes N/A

# Description

This PR adds some developer-only routes and utilities to aid with writing Cypress tests.

Routes Added:
- `GET /api/dev/login/:id?`
  If no `id` is provided, a **new** user is created and logged in.
- `POST /api/dev/run-util`
  Expects a body of `{ method: string; args: unknown[] }`.

Cypres Commands Added (aliases for the routes above):
- `createUsers(count?: number)`
- `login(id?: number)`
- `truncateDatabase()`
- `runUtil(method: string, ...args: unknown[])`
  This method can run any of the exported `/server/test/utils` functions. It accepts the method's name as the first parameter and the arguments to pass into the function as subsequent parameters.

Extras:
- `/cypress/e2e/dev.spec.ts` was added to demonstrate the use of the newly added commands & endpoint.

## Testing

Run automated tests, both Jest and Cypress.

## Type of change

- Other (test utilities)

## Screenshots

N/A

# Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18-alpine  
+FROM node:18-alpine
 
 WORKDIR /app
 
@@ -7,6 +7,8 @@ COPY . .
 RUN npm ci
 
 RUN npm run build
+
+RUN npm run migrate
 
 EXPOSE 3000
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18-alpine
+FROM node:18-alpine  
 
 WORKDIR /app
 
@@ -7,8 +7,6 @@ COPY . .
 RUN npm ci
 
 RUN npm run build
-
-RUN npm run migrate
 
 EXPOSE 3000
 
--- a/cypress/e2e/dev.spec.ts
+++ b/cypress/e2e/dev.spec.ts
@@ -2,6 +2,10 @@ import '../support/commands';
 
 // Example tests demonstrating use of the utility functions.
 describe('example login test', () => {
+  afterEach(() => {
+    cy.truncateDatabase();
+  });
+
   it('allows a user to login and logout', () => {
     cy.visit('http://localhost:3000');
     cy.contains('a', 'Log in').should('exist');
--- a/cypress/e2e/dev.spec.ts
+++ b/cypress/e2e/dev.spec.ts
@@ -1,6 +1,7 @@
 import '../support/commands';
 
-describe('example test', () => {
+// Example tests demonstrating use of the utility functions.
+describe('example login test', () => {
   it('allows a user to login and logout', () => {
     cy.visit('http://localhost:3000');
     cy.contains('a', 'Log in').should('exist');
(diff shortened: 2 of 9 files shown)


server/api/index.ts
```diff
@@ -1,11 +1,21 @@
...
 import userRouter from './users';
 
 const apiRouter = AsyncRouter();
 
+const env = cleanEnv(process.env, {});
+
 apiRouter.use('/users', userRouter);
 apiRouter.use('/auth', authRouter);
 
+if (!env.isProd) {
```
timmyichen, line 16: i think it's safer to use `env.isDev` here (that way it's _only_ set is NODE_ENV is development, whereas if it's unset for whatever reason, or set to something similar to but not exactly "production", this will still not apply)

server/api/dev/dev.controller.ts
timmyichen, line 1: For this file - I like the idea of having an explicit "login" method since that requires interacting with the response object - but for the others, what your thoughts on making a generic "run util" endpoint?  it would allow us to add additional test utils without having to maintain dev routes for them.  might look something like this:

```ts
import * as testUtils from "server/test/utils"

const postRunUtil: RequestHandler<some typings here> = async (req, res) => {
  const { method, args } = req.body;

  try {
    // @ts-ignore because we are doing typescript crimes here
    const result = await testUtils[method](args)
    res.json(result)
  } catch (err) {
    // handle error
  }
}
```

server/api/dev/dev.test.ts
timmyichen, line 1: doing the above (generic "run util" method) would also save us some time on writing tests for dev routes - would just need to test the login method + run util method

cypress/e2e/dev.spec.ts
```diff
@@ -0,0 +1,17 @@
+import '../support/commands';
+
+describe('example test', () => {
```
timmyichen, line 3: nit: "example test" - copy paste error?

timmyichen (approved):
Nice! I have some thoughts/opinions but nothing blocking. Thanks for doing this!

Dockerfile
```diff
@@ -8,6 +8,8 @@ RUN npm ci
 
 RUN npm run build
 
+RUN npm run migrate
```
timmyichen, line 11: i think migrations need to be run post-build as part of the GHA workflow, otherwise there's no db to connect to

devonzara: Ahh, you're right! I was thinking the `depends_on` would ensure that the database was running, I forgot that images are fetched/built before any container is started.

cypress/e2e/dev.spec.ts
devonzara: No, this was intentional... just a demonstration of how to login/logout using the added Cypress command.

Do you think we need tests to actually test the setup of the custom commands? 🤔

timmyichen: Haha fair enough! Probably not then

server/api/dev/dev.controller.ts
devonzara: Interesting, I definitely like the idea of doing things dynamically. Though, I guess my take on this would depend on if we anticipate the use of many other utility functions.

My understanding is that the idea behind E2E and Cypress is to do most things through the UI the way an actual user would. I think adding the ability to freely reach for too many utility functions may encourage reaching for them instead of writing tests that actually test the flow of the application.

That said, I'm still quite new to E2E testing, so... not sure how many other utility functions we may actually need. I'd say it's ultimately up to you (or others, if they wish to chime in).

timmyichen: Sure thing - and your understanding is on point.  The trouble becomes when we want to get a user into a specific state to start a test; getting a user into those states via the UI can be time consuming, if not straight up impossible (in the event that we deprecate some behavior but still want to support accounts in that legacy state).  Setting it up directly in the DB is the fastest and easiest way to do so.

devonzara:
@timmyichen Added a dynamic endpoint, as requested, and removed the other static endpoints. Please re-review.

timmyichen (approved):
🎉

windyScripts (approved):
LGTM 🎉

image

image

PR #108, opened 2023-06-14
#98 Notistack snackbars are functional and implemented through a custom hook. 

### LINKED ISSUE #98: Snackbar error handling
**Description**
We want to create a way for any part of the app to surface errors.  [Notistack](https://github.com/iamhosseindhv/notistack) does this pretty well.
* Alerts should support dismissing them, [example](https://notistack.com/features/basic#actions)
* It should support different kind of snackbars. Here are the [props](https://notistack.com/api-reference#mutual-props)

It would also be nice if we had a react hook that returns helper functions for enqueueing any success/error/warning/info.  So the actual usage would look something like this:

```ts
// within a react component
const { showSuccess, showError } = useAlerts();

const onClick = () => {
  try {
    doSomething();
    showSuccess('Thing was successful');
  } catch (err) {
    showError('something went wrong');
  }

  return <Button onClick={onClick}>Do a thing</Button>
}
```

**Acceptance Criteria**
* Notistack is used in the app
* Snackbars can be enqueued from any part of the app
* Abstractions exist via custom hook

windyScripts:
Closes #98 

# Description
created /client/components/SnackBar.tsx which returns a custom hook that in turn returns success and failure snackbars that take a message param and an optional params object.

## Testing

Within any component, calling the functions obtained from the useAlert custom hook with a message argument and an optional params object argument should create snackbars.

## Type of change

## Screenshots

Error Message

image

Success Message

image

# Checklist:

### DIFF
--- a/client/components/SnackBar.tsx
+++ b/client/components/SnackBar.tsx
@@ -1,0 +1,27 @@
+import { Button } from '@mui/material';
+import { useSnackbar, OptionsWithExtraProps } from 'notistack';
+import React from 'react';
+
+export function useAlerts(/* props: {  snackbarId: string; successMessage: string; errorMessage: string } */) {
+  const { enqueueSnackbar, closeSnackbar } = useSnackbar();
+  const action = (snackbarId: string) => (
+    <>
+      <Button onClick={() => {
+        closeSnackbar(snackbarId);
+      }}>
+        Dismiss
+      </Button>
+    </>
+  );
+
+  const successOptions: OptionsWithExtraProps<'success'> = { action, autoHideDuration: 6000, variant: 'success' };
+  const errorOptions: OptionsWithExtraProps<'error'> = { action, autoHideDuration: 6000, variant: 'error' };
+  return (
+    [(message: string, props?: OptionsWithExtraProps<'success'>) => {
+      enqueueSnackbar(message, { ...successOptions, ...props });
+    },
+    (message: string, props?: OptionsWithExtraProps<'error'>) => {
+      enqueueSnackbar(message, { ...errorOptions, ...props });
+    }]
+  );
+}
--- a/client/components/SnackBar.tsx
+++ b/client/components/SnackBar.tsx
@@ -17,11 +17,11 @@ export function useAlert() {
   const successOptions: OptionsWithExtraProps<'success'> = { action, autoHideDuration: 6000, variant: 'success' };
   const errorOptions: OptionsWithExtraProps<'error'> = { action, autoHideDuration: 6000, variant: 'error' };
   return (
-    [(message: string, props?: OptionsWithExtraProps<'success'>) => {
+    { showSuccess: (message: string, props?: OptionsWithExtraProps<'success'>) => {
       enqueueSnackbar(message, { ...successOptions, ...props });
     },
-    (message: string, props?: OptionsWithExtraProps<'error'>) => {
+    showError: (message: string, props?: OptionsWithExtraProps<'error'>) => {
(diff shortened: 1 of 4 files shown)


Caleb-Cohen:
I was able to implement this into a component pretty easily using the code below. How would I implement calling the success/failure while passing a variable message through?

```TS
  const { showSuccess, showError } = useAlert();

  const handleSuccess = () => {
    showSuccess(`Success Message`);
  };

  const handleError = () => {
    showError('Error message');
  };
  
    return (
        <Button onClick={handleSuccess} variant="contained" color="success">
        Show Success
      </Button>
      <Button onClick={handleError} variant="contained" color="error">
        Show Error
      </Button>
```

windyScripts:
>  How would I implement calling the success/failure while passing a variable message through?

I think we'd want each message triggered by a specific context, so instead of variable messages, we'd implement a message for each context that demands a message. This would become an issue if we do end up needing multiple messages for a single context or something similar.

Caleb-Cohen:
> > How would I implement calling the success/failure while passing a variable message through?
> 
> Good question. I think we'd want each message triggered by a specific context, so instead of variable messages, we'd implement a message for each context that demands a message. This would become an issue if we do end up needing multiple messages for a single context or something similar.

What if we wanted to pass back an error message that received as a response from the server?

jmoldyvan:
Would it be within the scope of the issue and beneficial to have other errors beyond success and failure, i.e. a warning?

timmyichen:
> I was able to implement this into a component pretty easily using the code below. How would I implement calling the success/failure while passing a variable message through?

Here's your example but with variable example (more realistically we'd use these with network requests to let them know if the thing they're trying to do succeeded or failed)

> ```ts
>   const { showSuccess, showError } = useAlert();
> 
>   const onClick = () => {
>     try {
>       await axios.post('/some-route');
>       showSuccess("Completed job");
>     } catch (err) {
>       showError(err.response?.data?.message || "Could not complete job");
>     }
>   }
>   
>     return (
>         <Button onClick={handleSuccess} variant="contained" color="success">
>         Show Success
>       </Button>
>       <Button onClick={handleError} variant="contained" color="error">
>         Show Error
>       </Button>
> ```

client/components/SnackBar.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const successOptions: OptionsWithExtraProps<'success'> = { action, autoHideDuration: 6000, variant: 'success' };
+  const errorOptions: OptionsWithExtraProps<'error'> = { action, autoHideDuration: 6000, variant: 'error' };
+  return (
+    { showSuccess: (message: string, props?: OptionsWithExtraProps<'success'>) => {
+      enqueueSnackbar(message, { ...successOptions, ...props });
+    },
+    showError: (message: string, props?: OptionsWithExtraProps<'error'>) => {
+      enqueueSnackbar(message, { ...errorOptions, ...props });
+    } }
+  );
```
timmyichen, line 26: nit: styling - better for return values to be as simple as possible to make it really obvious what's being returned from this hook.

```suggestion
  const showSuccess = (message: string, props?: OptionsWithExtraProps<'success'>) => {
    enqueueSnackbar(message, { ...successOptions, ...props });
  };
  const showError = (message: string, props?: OptionsWithExtraProps<'error'>) => {
    enqueueSnackbar(message, { ...errorOptions, ...props });
  };
  
  return {
    showSuccess,
    showError,
  };
```

timmyichen (approved):
Nice work!  Looks good to me! :)

Caleb-Cohen (approved):

